PR #10, opened 2021-09-09
Allow persisting plugin state (reverb tails, etc) between calls to render() if reset=False.

psobot:
Pedalboard currently clears any internal plugin state by calling `Plugin::reset` every time it processes a new buffer of audio. This is great for the batch use case (as it keeps output consistent) but prevents Pedalboard from being used to process audio incrementally. (i.e.: adding reverb to a buffer in chunks is currently not possible, as the state inside the reverb plugin would be reset between each chunk of audio, cutting off the reverb tail.)

This PR adds a new `reset: bool` flag, which defaults to `True`. If set to `False`, Pedalboard will avoid resetting the plugin instance between renders, which allows its usage on chunks of audio (or even in real-time). This is risky, however, as the same plugin object can't be used in multiple pedalboards if `reset=False`.

After this change, it's now possible to process audio even without loading the whole file into memory with something like:
```python3
import soundfile as sf
from pedalboard import Pedalboard, Compressor, Reverb

with sf.SoundFile("./my-hour-long-audio.wav") as input_file, sf.SoundFile(
    "./output.wav",
    "w",
    samplerate=input_file.samplerate,
    channels=input_file.channels
) as output_file:
    board = Pedalboard([Compressor(), Reverb()], sample_rate=input_file.samplerate)

    # Only load and process 1,024 samples at a time to keep memory usage super low:
    for block in input_file.blocks(1024):
        output_file.write(board.process(block, reset=False)) # reset=False is crucial here!
```

This PR also includes a tiny optimization: `prepare` is no longer called on each plugin unless the number of input channels, sample rate, or size of each sample buffer changes. This should remove a tiny amount of overhead from successive calls to `process`, as plugins no longer need to be re-initialized.

### DIFF
--- a/examples/add_reverb_to_file.py
+++ b/examples/add_reverb_to_file.py
@@ -1,0 +1,132 @@
+"""
+Add reverb to an audio file using Pedalboard.
+
+The audio file is read in chunks, using nearly no memory.
+This should be one of the fastest possible ways to add reverb to a file.
+On my laptop, this runs about 32x faster than real-time (i.e.: processes a 60-second file in <2 seconds.)
+"""
+
+import sys
+import warnings
+import argparse
+import numpy as np
+from tqdm import tqdm
+import soundfile as sf
+from tqdm.std import TqdmWarning
+from pedalboard import Reverb
+
+
+BUFFER_SIZE_SAMPLES = 1024
+NOISE_FLOOR = 1e-4
+
+
+def get_num_frames(f: sf.SoundFile) -> int:
+    # On some platforms and formats, f.frames == -1L.
+    # Check for this bug and work around it:
+    if f.frames > 2 ** 32:
+        f.seek(0)
+        last_position = f.tell()
+        while True:
+            # Seek through the file in chunks, returning
+            # if the file pointer stops advancing.
+            f.seek(1024 * 1024 * 1024, sf.SEEK_CUR)
+            new_position = f.tell()
+            if new_position == last_position:
+                return new_position
+            else:
+                last_position = new_position
+    else:
+        return f.frames
+
+
+def main():
+    warnings.filterwarnings("ignore", category=TqdmWarning)
+
+    parser = argparse.ArgumentParser(description="Add reverb to an audio file.")
+    parser.add_argument("input_file", help="The input file to add reverb to.")
+    parser.add_argument(
+        "--output-file",
+        help=(
+            "The name of the output file to write to. If not provided, {input_file}.reverb.wav will"
+            " be used."
+        ),
+        default=None,
+    )
+
+    # Instantiate the Reverb object early so we can read its defaults for the argparser --help:
+    reverb = Reverb()
+
+    parser.add_argument("--room-size", type=float, default=reverb.room_size)
+    parser.add_argument("--damping", type=float, default=reverb.damping)
+    parser.add_argument("--wet-level", type=float, default=reverb.wet_level)
+    parser.add_argument("--dry-level", type=float, default=reverb.dry_level)
+    parser.add_argument("--width", type=float, default=reverb.width)
+    parser.add_argument("--freeze-mode", type=float, default=reverb.freeze_mode)
+
+    parser.add_argument(
+        "--cut-reverb-tail",
+        action="store_true",
+        help=(
+            "If passed, remove the reverb tail to the end of the file. "
+            "The output file will be identical in length to the input file."
+        ),
+    )
+    args = parser.parse_args()
+
+    for arg in ('room_size', 'damping', 'wet_level', 'dry_level', 'width', 'freeze_mode'):
+        setattr(reverb, arg, getattr(args, arg))
+
+    if not args.output_file:
+        args.output_file = args.input_file + ".reverb.wav"
+
+    sys.stderr.write(f"Opening {args.input_file}...\n")
+
+    with sf.SoundFile(args.input_file) as input_file:
+        sys.stderr.write(f"Writing to {args.output_file}...\n")
+        with sf.SoundFile(
+            args.output_file,
+            'w',
+            samplerate=input_file.samplerate,
+            channels=input_file.channels,
+        ) as output_file:
+            length = get_num_frames(input_file)
+            length_seconds = length / input_file.samplerate
+            sys.stderr.write(f"Adding reverb to {length_seconds:.2f} seconds of audio...\n")
+            with tqdm(
+                total=length_seconds,
+                desc="Adding reverb...",
+                bar_format=(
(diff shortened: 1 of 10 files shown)


pedalboard/ExternalPlugin.h
```diff
@@ -222,6 +222,13 @@ template <typename ExternalPluginType> class ExternalPlugin : public Plugin {
 
     pluginInstance->setStateInformation(savedState.getData(),
                                         savedState.getSize());
+    
+    const juce::dsp::ProcessSpec _lastSpec = lastSpec;
```
drubinstein, line 226: This doesn't haven't to be in the lock guard, correct?

tests/test_native_module.py
```diff
@@ -89,3 +89,30 @@ def test_distortion(gain_db, shape, sr=44100):
...
+    state should be maintained between calls to `render`
+    (preserving tails).
+    """
+    reverb = Reverb()
+    sr = 44100
+    noise = np.random.rand(sr)
+    silence = np.zeros_like(noise)
+
+    # Assert that reverb adds nothing if no signal was present already:
+    assert np.amax(np.abs(reverb(silence, sr, reset=reset))) == 0.0
```
drubinstein, line 108: I'd consider making this check that the amax is < some small value like 1e-5 to avoid floating point comparison problems, but 0.0 shouldnt be an issue.

drubinstein (approved):
Looks like there are some clang format errors that need to be addressed, but the rest seems reasonable. Maybe it's overkill, but going forward you could add some githooks for running clang format on commit + adding a .clangformat file to the repo.

tests/test_native_module.py
psobot: Agreed - luckily the built-in `Reverb` plugin is numerically stable and doesn't add any noise/dither, so this assertion passes. Other plugins or VSTs would probably fail this check.

pedalboard/ExternalPlugin.h
psobot: Correct - Python can only ever reach this code by calling through `process()`, at which point it'll [already be holding a `scoped_lock`](https://github.com/spotify/pedalboard/blob/master/pedalboard/process.h#L149-L153) on the entire `Plugin` instance.

examples/add_reverb_to_file.py
```diff
@@ -0,0 +1,135 @@
...
+from pedalboard import Reverb
+
+
+BUFFER_SIZE_SAMPLES = 1024
+NOISE_FLOOR = 1e-4
+
+
+def get_num_frames(f: sf.SoundFile) -> int:
+    # On some platforms and formats, f.frames == -1L.
+    # Check for this bug and work around it:
```
drubinstein, line 28: This is the Mac OS bug right? You have a link to the GH issue?

examples/add_reverb_to_file.py
```diff
@@ -0,0 +1,135 @@
...
+
+The audio file is read in chunks, using nearly no memory.
+This should be one of the fastest possible ways to add reverb to a file.
+On my laptop, this runs about 32x faster than real-time
+(i.e.: processes a 60-second file in <2 seconds.)
+
+Requirements: `pip install pysoundfile tqdm pedalboard`
+"""
+
+import sys
```
drubinstein, line 12: Probably should be isorted 🤷

examples/add_reverb_to_file.py
```diff
@@ -0,0 +1,135 @@
...
+
+    # Instantiate the Reverb object early so we can read its defaults for the argparser --help:
+    reverb = Reverb()
+
+    parser.add_argument("--room-size", type=float, default=reverb.room_size)
+    parser.add_argument("--damping", type=float, default=reverb.damping)
+    parser.add_argument("--wet-level", type=float, default=reverb.wet_level)
+    parser.add_argument("--dry-level", type=float, default=reverb.dry_level)
+    parser.add_argument("--width", type=float, default=reverb.width)
+    parser.add_argument("--freeze-mode", type=float, default=reverb.freeze_mode)
```
drubinstein, line 67: Are there limits on the floats that should be enforced?

examples/add_reverb_to_file.py
```diff
@@ -0,0 +1,135 @@
...
+    else:
+        return f.frames
+
+
+def main():
+    warnings.filterwarnings("ignore", category=TqdmWarning)
+
+    parser = argparse.ArgumentParser(description="Add reverb to an audio file.")
+    parser.add_argument("input_file", help="The input file to add reverb to.")
+    parser.add_argument(
```
drubinstein, line 50: Possible gotcha but should there be any checks to make sure the file doesn't exist/a flag to disable that or something like add _# to the file if the same file is found?

examples/add_reverb_to_file.py
```diff
@@ -0,0 +1,135 @@
...
+                    # continue from one chunk to the next.)
+                    effected_chunk = reverb.process(
+                        dry_chunk, sample_rate=input_file.samplerate, reset=False
+                    )
+                    # print(effected_chunk.shape, np.amax(np.abs(effected_chunk)))
+                    output_file.write(effected_chunk)
+                    t.update(len(dry_chunk) / input_file.samplerate)
+                    t.refresh()
+            if not args.cut_reverb_tail:
+                while True:
```
drubinstein, line 121: An alternative condition idea - pull until you match the number of bytes input. A little fear I have with the break condition you set is that what if a user intentionally inputs silence?

drubinstein (commented):
Decided to review the add_reverb_to_file for fun. I'm probably over-engineering the script, but I did recently run into an issue where an example script was overwriting files without telling us after resampling so...

christinesutcliffe:
> cc @christinesutcliffe - this PR adds an `examples` folder. I know you said you'd be interested in adding some examples - there's now a home for them!

Amazing! Thanks for setting this up, I'll make sure to add my files here!

examples/add_reverb_to_file.py
psobot: I might have been able to name this a bit more clearly - the loop before this `if` statement does exactly what you suggest (pull until you match the number of bytes input). This just extends the output file until the effect processor has stopped making any sound.

examples/add_reverb_to_file.py
psobot: There are, but the setters on the `Reverb` instance will throw descriptive `ValueErrors` if the floats are out of range.

examples/add_reverb_to_file.py
psobot: I'll add an `-y/--overwrite` flag like `ffmpeg` has. 👍🏻

PR #15, opened 2021-09-12
Add --recurse-submodules to git clone in CONTRIBUTING.md

rec:
The instructions as written don't check out the `JUCE` subtree, so I tweaked 'em.

I also added `--shallow-submodules` because that speeds up checking out the huge JUCE module considerably, though that step is still rather slow.


rec:
Anytime!  See also [this](https://www.reddit.com/r/pedalboard/comments/pmtim9/my_first_impressions_of_pedalboard/?)

PR #22, opened 2021-09-15
Add `pedalboard.__version__`

reshalfahsi:
Add __version__ to check the current version of the package
Example:

```python
import pedalboard
print(pedalboard.__version__)
```

The output should be

    '0.3.6'


saveriogzz (changes requested):
Hey @reshalfahsi ! I reckon you should also change the hard-coded `version` variable in the `setup.py` file.

I guess that, according to PEP8, we shall create a file called `version.py` into the pedalboard folder with, for instance, the following content:

```python
MAJOR = 0
MINOR = 3
PATCH = 6
__version__ = '%d.%d.%d' % (MAJOR, MINOR, PATCH)
```

and in `setup.py` we'll have something like
```python
# read version
version = {}
with open("pedalboard/version.py") as fp:
    exec(fp.read(), version)

setup(
    name='pedalboard',
    version=version['__version__'],
    author='Peter Sobot',
[....]
```
Source: [https://packaging.python.org/guides/single-sourcing-package-version/]()

reshalfahsi:
How about this?

saveriogzz:
> How about this?

Try to stick to PEP as described above! :)
If you're okay with it, just click on the `Allow edits from mainteners.` on the right hand side of this PR view; I shall then be able to push my edits here!

Cheers

saveriogzz:
I cannot do it - most likely because I'm not a maintainer.
Try to just follow the instructions in my comment above and commit once more. If you can't, I will open another PR and refer to this one - but let's try not to pollute the project with too many PRs :beers:

reshalfahsi:
> I cannot do it - most likely because I'm not a maintainer.
> Try to just follow the instructions in my comment above and commit once more. If you can't, I will open another PR and refer to this one - but let's try not to pollute the project with too many PRs

Ok, I'll try to do that just be patient with me :)
Just comment where I should change later :smile:

pedalboard/__init__.py
```diff
@@ -20,3 +20,6 @@
 
 for klass in AVAILABLE_PLUGIN_CLASSES:
     vars()[klass.__name__] = klass
+
+# pedalboard current version
+__version__ = '0.3.6'
```
saveriogzz: let's clean out this!

pedalboard/version.txt
```diff
@@ -0,0 +1 @@
+0.3.6
```
saveriogzz, line 1: We might want this file to have a `py` extension.
For the content, we could insert something like:

```python
MAJOR = 0
MINOR = 3
PATCH = 6
__version__ = '%d.%d.%d' % (MAJOR, MINOR, PATCH)
```

pedalboard/__init__.py
reshalfahsi: I have. Have you seen my latest change?

setup.py
```diff
@@ -202,9 +202,13 @@
 this_directory = Path(__file__).parent
 long_description = (this_directory / "README.md").read_text()
 
+# current version
+with open(os.path.join("pedalboard", "version.txt")) as f:
+    version = f.read().strip()
+
```
saveriogzz, line 208: Sticking to PEP documentation, let's do something like this:

```python
# read version
version = {}
with open("pedalboard/version.py") as fp:
    exec(fp.read(), version)
```

setup.py
```diff
@@ -202,9 +202,13 @@
...
 setup(
     name='pedalboard',
-    version='0.3.6',
+    version=version,
```
saveriogzz, line 211: So that here will be `version=version['__version__'],`

pedalboard/version.txt
reshalfahsi: So, the `__version__` should be placed in the separated file?

setup.py
reshalfahsi: I see, working on it

pedalboard/version.txt
saveriogzz: rename the file from `version.txt` to `version.py` and place the content I wrote above

setup.py
```diff
@@ -226,5 +230,6 @@
     ],
     ext_modules=[pedalboard_cpp],
     install_requires=['numpy'],
+    package_data={'pedalboard': ['version.txt']},
```
saveriogzz, line 233: this can be removed as well

pedalboard/__init__.py
saveriogzz: no need for `__version__` to be initialized here, you can do it in `version.py`

reshalfahsi:
Let see my latest commit. How is it?

saveriogzz (approved):
Cool! Thanks :beers:

reshalfahsi:
> Cool! Thanks

At last, I understood what you mean, haha :laughing:

setup.py
```diff
@@ -202,9 +202,14 @@
 this_directory = Path(__file__).parent
 long_description = (this_directory / "README.md").read_text()
 
+# read version
+version = {}
+with open("pedalboard/version.py") as fp:
+    exec(fp.read(), version)
```
psobot, line 208: We use [the `pathlib` library](https://docs.python.org/3/library/pathlib.html) just a couple lines above to do a very similar task. What about using that here too?

```suggestion
version_file_contents = (this_directory / "pedalboard" / "version.py").read_text()
exec(version_file_contents, version)
```

psobot (approved):
Thanks for this! This looks great. I've left a couple of suggestions - happy to merge once they're resolved.

reshalfahsi:
Hi @psobot, please take a look at my new commit. I did just like the suggestions. Thanks for the suggestions.

psobot (approved):

PR #46, opened 2021-11-20
feature: set_raw_value

carlostorreswav:
**module-name:** feature to set raw float values of plugin parameters

**Problem:** There may be cases where you are reading float values (from a knob, for example) and I think it could be useful to be able to set directly the raw value of a specific parameter.

**Solution:** a `set_raw_value` function that receives the parameter name and the raw_value to set it.

**Result:** e.g., `my_plugin.set_raw_value('gain', float_value)`

### DIFF
--- a/pedalboard/pedalboard.py
+++ b/pedalboard/pedalboard.py
@@ -572,6 +572,22 @@ class ExternalPlugin(object):
             )
         return self.__python_parameter_cache__[cpp_parameter.name]
 
+    def set_raw_value(self, name: str, raw_value: float):
+        """
+        Function to set raw value for a given parameter
+        :param name: Name of the parameter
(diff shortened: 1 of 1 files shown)


carlostorreswav:
Hi @psobot, thank you for the quick response.

I thought raw_value was gettable but not settable because an `'VST3Plugin' object is not subscriptable` error, but was my mistake because `my_plugin.parameters[key].raw_value = float_value` indeed works well, sorry for the misunderstanding 😅

I close this PR as this feature is already implemented.

Btw, I'm developing a (little) latency-compensation feature because very high-cpu usage plugins (e.g., FabFilter) have a noticeable latency (over 80ms). I may let you know if I achieve a clean workaround, if only for curiosity's sake

Have a very nice day.

PR #50, opened 2021-12-16
Add frequently asked questions to the README.

psobot:


README.md
```diff
@@ -180,6 +180,58 @@ For more examples, see:
...
 Contributions to `pedalboard` are welcomed! See [CONTRIBUTING.md](https://github.com/spotify/pedalboard/blob/master/CONTRIBUTING.md) for details.
 
+## Frequently Asked Questions
+
+
+### Can Pedalboard be used with live (real-time) audio?
+
+Technically, yes, Pedalboard could be used with live audio input/output. See [@stefanobazzi](https://github.com/stefanobazzi)'s [guitarboard](https://github.com/stefanobazzi/guitarboard) project for an example that uses the `python-sounddevice` library to wire Pedalboard up to live audio.
+
+However, there are a couple big caveats when talking about using Pedalboard in a live context. Python, as a language, is [garbage-collected](https://devguide.python.org/garbage_collector/), meaning that your code randomly pauses on a regular interval to clean up unused objects. In most programs, this is not an issue at all. However, for live audio, garbage collection can result in random pops, clicks, or audio drop-outs that are very difficult to prevent.
```
drubinstein, line 190: You could pause the garbage collector though?

Another caveat that may be worth mentioning is that the [overhead](https://pythonspeed.com/articles/python-extension-performance/) from serialization/deserialization to/from the C++-bindings.

README.md
```diff
@@ -180,6 +180,58 @@ For more examples, see:
...
+
+With this technique, it's possible to automate any parameter. Usually, using a step size of somewhere between 100 and 1,000 (2ms to 22ms at a 44.1kHz sample rate) is small enough to avoid hearing any audio artifacts, but big enough to avoid slowing down the code dramatically.
+
+### Can Pedalboard be used with VST instruments, instead of effects?
+
+Not yet! The underlying framework (JUCE) supports VST and AU instruments just fine, but Pedalboard itself would have to be modified to support instruments.
+
+### Can Pedalboard plugins accept MIDI?
+
+Not yet, either - although the underlying framework (JUCE) supports passing MIDI to plugins, so this would also be possible to add.
```
drubinstein, line 233: Regarding usage of "not yet" do we want to use the word "yet" as it implies there is a plan to support it in the future? Do we expect to do it? I think "Not planned" or some phrasing indicating lack of desire to support and write it may be a good idea.

Also, scope creep.

drubinstein (approved):

README.md
psobot: I suppose I should add more detail here - even with the garbage collector paused, other unpredictable delays can result from [the Python GIL](https://wiki.python.org/moin/GlobalInterpreterLock) being taken automatically during audio callbacks, which is necessary as long as Python is being called from the audio thread. Basically the only way to process live audio with Pedalboard without high latency or risk of buffer underruns would be to cut Python out of the audio thread entirely, which would require tight integration with the C++ side of Pedalboard. (For more on this, I love [this talk by **@timuraudio** from CppCon 2015](https://www.youtube.com/watch?v=boPEO2auJj4).)

Also, re ser/de overhead: luckily the vast majority of the data being passed from Python to C++ (and vice versa) is in a single, large NumPy array, which makes that overhead pretty much negligible.

README.md
psobot: Definite scope creep! I said "not yet" here as I expect either I (or someone else) will eventually add MIDI input support, as it's desirable, possible (fairly easily), but not explicitly planned.

Issue #11, opened 2021-09-11
Is there a way to open preset file?


WonGilHuh:
I am using an external VST3 Plugin and MacOS.

Is there any way to load .fxp preset file or .patch file?
I want to use parameters stored in the preset file.

psobot:
Great question! There isn't currently a way to do this, but I see that JUCE itself offers a method for it: [`setStateFromVSTPresetFile`](https://github.com/juce-framework/JUCE/blob/fbe95b0b073ccb589e28255cb0cd56e1382dc2c9/modules/juce_audio_processors/format_types/juce_VST3PluginFormat.h#L50). This should be fairly easy to add - I'll leave this issue open to track this feature request.

psobot:
Hi @HussainAbdi! This would be a great first issue, yup - I haven't played around with the new `getExtensions` API, but that looks like the way forward. That might require an update to the version of JUCE used, which could be a pain, but worth trying.

HussainAbdi:
Good evening @psobot! Finals are almost over so I'm excited to finally (pun intended, forgive me) make some progress on this issue. I was wondering; why would updating JUCE be a pain? Would it not be as simple as updating the submodule or are there breaking changes or other things I am unaware of? I'm also running into some issues when running the `tox` command, but I'll open up another "question" issue regarding that. Hope you're well!

psobot:
@HussainAbdi It might be that simple! Hard to tell, unfortunately - I'm not sure what's changed in JUCE between the version currently being used in Pedalboard ([6.0.8](https://github.com/juce-framework/JUCE/releases/tag/6.0.8)) and the latest ([6.1.3](https://github.com/juce-framework/JUCE/releases/tag/6.1.3)). Luckily, we have enough tests that any breaking changes should be caught automatically.

simonbowen:
@HussainAbdi I am absolutely desperate for this functionality. Currently using mrswatson for some headless VST processing and this would be an ideal replacement. I would love to help in anyway but have no C++ experience.

HussainAbdi:
@simonbowen It's good to see people need this! This is my first issue for any open source project so I'm afraid you'll have to be patient a little bit longer while I work on this. I've been documenting everything I'm doing so as soon as I get a grasp on exactly what it is I need to do, I'll let you know and fill you in. Happy holidays!

simonbowen:
@HussainAbdi No worries. I just was I had to the skillset to be of more use!

psobot:
[...]

PR #55, opened 2021-12-28
Updated JUCE to version 6.1.4

HussainAbdi:
update-juce: Checked out the latest version of JUCE

Problem
In order to work on [this issue](https://github.com/spotify/pedalboard/issues/11), an update of JUCE is first needed. The discussion there outlines why. 

Solution
I simply updated the JUCE submodule to the latest stable master commit (their commit message is aptly "JUCE version 6.1.4"). I made sure to detach the HEAD in order to prevent automatically updating in the future. 

Comments
Running tox, the tests on my end seem to be all good. I was unable to run lint tests but this isn't a concern for this specific PR. Lint seems to be concerned with syntactical and stylistic problems but the only thing this PR is doing is updating JUCE. Moving forward I will need to sort out this lint issue.

P.S. This is my first PR to any open source project! It's quite a trivial one, but I look forward to contributing more. Let me know if I did anything wrong and thank you for your patience with my learning in advance.

### DIFF
--- a/pedalboard/JuceHeader.h
+++ b/pedalboard/JuceHeader.h
@@ -32,7 +32,7 @@
 #include <juce_graphics/juce_graphics.h>
 #include <juce_gui_basics/juce_gui_basics.h>
 
-#define JUCE_PROJUCER_VERSION 0x60008
+#define JUCE_PROJUCER_VERSION 0x60104
 
 #if JUCE_PROJUCER_VERSION < JUCE_VERSION
 /** If you've hit this error then the version of the Projucer that was used to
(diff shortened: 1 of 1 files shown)


HussainAbdi:
Hi @psobot! That makes sense, but if pedalboard isn't using the Projucer could we instead remove [these lines](https://github.com/spotify/pedalboard/blob/4bb35773e78bd14c006d97b7386d3bbd7e777a2f/pedalboard/JuceHeader.h#L35-L46) altogether?

HussainAbdi:
Done! That's great to hear. You even removed the tensorflow test for M1 Macs. Thank you for that, I was doing that manually on my end. Also, congratulations on 3000 stars!

PR #56, opened 2021-12-30
Fix tests when running on Apple Silicon.

### LINKED ISSUE #32: All the tests involving CHOWTapeModel.component fail
I have create a conda environment, installed `pybind11` and `tox` via `conda` and run `tox`. But all the tests involving `CHOWTapeModel.component` fail.

Some more details (please tell me if you want more information):
- macOS BigSur 11.4
- Python 3.8.12

Any idea?

psobot:
Fixes #32 and gets tests running properly on Apple Silicon. Verified on an M1 MacBook Air with macOS 11.4 (Big Sur).

### DIFF
--- a/pedalboard/ExternalPlugin.h
+++ b/pedalboard/ExternalPlugin.h
@@ -194,9 +194,8 @@ public:
           "see which dependencies might be missing.).");
 #else
 
-// Only on M1 Macs, it seems that plugins must be installed in the appropriate
-// path.
-#if JUCE_ARM && JUCE_MAC
+// On certain Macs, plugins will only load if installed in the appropriate path.
+#if JUCE_MAC
       bool pluginIsInstalled =
           pluginFileStripped.contains("/Library/Audio/Plug-Ins/Components/");
       if (!pluginIsInstalled) {
--- a/pedalboard/ExternalPlugin.h
+++ b/pedalboard/ExternalPlugin.h
@@ -193,6 +193,20 @@ public:
           pathToSharedObjectFile.getFullPathName().toStdString() + "\"` to " +
           "see which dependencies might be missing.).");
 #else
+
+// Only on M1 Macs, it seems that plugins must be installed in the appropriate
+// path.
+#if JUCE_ARM && JUCE_MAC
+      bool pluginIsInstalled =
+          pluginFileStripped.contains("/Library/Audio/Plug-Ins/Components/");
+      if (!pluginIsInstalled) {
+        throw pybind11::import_error(
+            "Unable to load plugin " + pathToPluginFile.toStdString() +
+            ": unsupported plugin format or load failure. Plugin file may need "
+            "to be moved to /Library/Audio/Plug-Ins/Components/ or "
+            "~/Library/Audio/Plug-Ins/Components/.");
+      }
+#endif
       throw pybind11::import_error(
           "Unable to load plugin " + pathToPluginFile.toStdString() +
           ": unsupported plugin format or load failure.");
--- a/tests/test_external_plugins.py
+++ b/tests/test_external_plugins.py
@@ -76,7 +76,7 @@ def load_test_plugin(plugin_filename: str, disable_caching: bool = False, *args,
         plugin_path = os.path.join(TEST_PLUGIN_BASE_PATH, platform.system(), plugin_filename)
 
         if (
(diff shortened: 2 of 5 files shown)


pedalboard/ExternalPlugin.h
```diff
@@ -193,6 +193,19 @@ template <typename ExternalPluginType> class ExternalPlugin : public Plugin {
           pathToSharedObjectFile.getFullPathName().toStdString() + "\"` to " +
           "see which dependencies might be missing.).");
 #else
+
+// On certain Macs, plugins will only load if installed in the appropriate path.
+#if JUCE_MAC
```
drubinstein, line 198: Since this exists in a function you can use [if constexpr](https://en.cppreference.com/w/cpp/language/if) instead. Should be safer than relying on the C macro preprocessor.

tests/test_external_plugins.py
```diff
@@ -87,6 +122,14 @@ def load_test_plugin(plugin_filename: str, disable_caching: bool = False, *args,
     return plugin
 
 
+def delete_installed_plugins():
+    for plugin_file in PLUGIN_FILES_TO_DELETE:
+        shutil.rmtree(plugin_file)
+
+
+atexit.register(delete_installed_plugins)
```
drubinstein, line 130: Looks like the more "pytest" way of doing this is to make this a [fixture](https://docs.pytest.org/en/6.2.x/fixture.html#autouse-fixtures-fixtures-you-don-t-have-to-request) in conftest.py with` scope="session"`  and `autouse=true` then pass it to every test. They recommend using [yield for clean up](https://docs.pytest.org/en/6.2.x/fixture.html#yield-fixtures-recommended).

With this I think you can install the plugin file copying/deletion as different fixtures.

tests/test_tensorflow.py
```diff
@@ -28,8 +29,10 @@ def test_can_be_called_in_tensorflow_data_pipeline():
     try:
         import tensorflow as tf
     except ModuleNotFoundError:
-        # TensorFlow is not yet supported on Python 3.10 - don't bother testing.
-        if sys.version_info.minor >= 10:
+        # TensorFlow is not yet supported on Python 3.10 or M1 Macs - don't bother testing.
+        if sys.version_info.minor >= 10 or (
```
drubinstein, line 33: Maybe a warning or equivalent should be logged to point out that a test on an unsupported platform was attempted?

tests/test_external_plugins.py
```diff
@@ -43,15 +46,21 @@
...
 TEST_PLUGIN_CACHE = {}
 TEST_PLUGIN_ORIGINAL_PARAMETER_CACHE = {}
 
+# If true, copy all test Audio Units into the appropriate install path
+# before running tests. On some versions of macOS, this is necessary or
+# else Audio Units won't load.
+TEMPORARILY_INSTALL_AUDIO_UNITS = True
+PLUGIN_FILES_TO_DELETE = set()
+MACOS_PLUGIN_INSTALL_PATH = os.path.expanduser(
+    os.path.join("~", "Library", "Audio", "Plug-Ins", "Components")
```
drubinstein, line 61: [Path.home](https://docs.python.org/3/library/pathlib.html#pathlib.Path.home) is the more direct way of getting the home dir and is available for python >3.5

pedalboard/ExternalPlugin.h
psobot: Unfortunately that won't work here - [`JUCE_MAC` (and `JUCE_LINUX`, etc) is only defined in the first place](https://github.com/juce-framework/JUCE/blob/4f01392000965f6c5e957329220d5ad06ca174c4/modules/juce_core/system/juce_TargetPlatform.h#L59-L84) if we're running on that platform, and otherwise it's not defined at all.

tests/test_tensorflow.py
psobot: I'll use [`pytest.mark.skipif`](https://docs.pytest.org/en/latest/how-to/skipping.html#id1) here instead.

tests/test_external_plugins.py
psobot: I gave that a try earlier - I'm pretty sure it could work as well, but requires moving this logic to `conftest.py` (which spreads this functionality across multiple files) and makes it much less clear when this is happening.

PR #59, opened 2022-01-18
Add PitchShift plugin.

Johnxjp:
Add pitch shift plugin to Pedalboard using the Rubberband library for processing the audio:
- Modifies the pitch without changing the duration of the audio. Pitch can be adjusted with a scale factor
- Adds the Rubberband source code to pedalboard
- Adds a base Rubberband plugin that can be extended for time stretching


pedalboard/RubberbandPlugin.h
```diff
@@ -0,0 +1,93 @@
...
+          outBlock[c] += missingSamples;
+        }
+      }
+
+      // Pull the next audio data out of Rubberband:
+      rbPtr->retrieve(outBlock, samplesToPull);
+    }
+
+  protected:
+    RubberBandStretcher *rbPtr = nullptr;
```
psobot, line 91: Should probably use a [`std::unique_ptr`](https://www.cplusplus.com/reference/memory/unique_ptr/) here so that this pointer is automatically cleaned up once this plugin is deleted. (As it stands now, this will leak a `RubberBandStretcher` instance every time this plugin is instantiated.)

```suggestion
    std::unique_ptr<RubberBandStretcher> rbPtr;
```

pedalboard/plugins/PitchShift.h
```diff
@@ -0,0 +1,78 @@
...
+    {
+    private:
+        double _pitchScale = 1.0;
+
+    public:
+        void setPitchScale(double scale)
+        {
+            if (scale <= 0)
+            {
+                throw std::range_error("Pitch scale be a value > 0");
```
psobot, line 40:
```suggestion
                throw std::range_error("Pitch scale must be a value greater than 0.0.");
```

pedalboard/plugins/PitchShift.h
```diff
@@ -0,0 +1,78 @@
...
+            }
+            _pitchScale = scale;
```
psobot, line 42: I think we'll also need to pass this `scale` value on to the `RubberBandStretcher` here, if one exists:
```suggestion
            _pitchScale = scale;
            if (rbPtr) rbPtr->setPitchScale(_pitchScale);
```

pedalboard/plugins/PitchShift.h
```diff
@@ -0,0 +1,78 @@
...
+
+            if (!rbPtr || specChanged)
+            {
+                auto stretcherOptions = RubberBandStretcher::OptionProcessRealTime | RubberBandStretcher::OptionThreadingNever;
+                auto rb = new RubberBandStretcher(spec.sampleRate, spec.numChannels, stretcherOptions);
+                rb->setMaxProcessSize(spec.maximumBlockSize);
+                rb->setPitchScale(_pitchScale);
+                rb->reset();
+                delete rbPtr;
+                rbPtr = rb;
```
psobot, line 61: If using `std::unique_ptr`, we'll likely want to change this to:
```suggestion
                rbPtr = std::make_unique<RubberBandStretcher>(spec.sampleRate, spec.numChannels, stretcherOptions);
                rbPtr->setMaxProcessSize(spec.maximumBlockSize);
                rbPtr->setPitchScale(_pitchScale);
                rbPtr->reset();
```

psobot (commented):
Looks great so far! I've added a couple comments around modern C++ tips to avoid dealing with raw pointers.

We might also want to consider making RubberBand a submodule before this work gets merged, as its [source is now available on GitHub](https://github.com/breakfastquay/rubberband) and would help us avoid adding 22,468 lines of library code along with this pull request. 🙂

Re: the issue of latency added by Rubberband; unfortunately I think we might have to live with it for now, as Rubberband's internal buffering of audio seems to require a certain amount of overlap between audio segments. (To put it another way: we could remove the delay from the start of each chunk processed by Rubberband, but then Rubberband wouldn't return us enough audio to fill the same-sized buffer that it accepted, as it'd be holding some of the audio in its internal buffer to combine with the next chunk we pass in.)

I might consider adding a `latency` read-only parameter to Pedalboard plugins to allow users to compensate for this later, but it'll take a bit of effort to think through the edge cases in the API.

pedalboard/RubberbandPlugin.h
```diff
@@ -0,0 +1,87 @@
...
+      auto outBlock = context.getOutputBlock();
+
+      auto len = inBlock.getNumSamples();
+      auto numChannels = inBlock.getNumChannels();
+
+      jassert(len == outBlock.getNumSamples());
+      jassert(numChannels == outBlock.getNumChannels());
+
+      const float **inChannels =
+          (const float **)alloca(numChannels * sizeof(float *));
```
drubinstein, line 28: Any particular reason as to why you're using alloca and not a `std::vector<float*>`  or a float[][]? alloca allocates on the stack and that can be dangerous in some cases.

But also do you know if the samples or channels will be constant throughout the use of this plugin? If so you should probably preallocate this in the constructor.

pedalboard/RubberbandPlugin.h
```diff
@@ -0,0 +1,87 @@
...
+    // If we don't have enough samples to fill a full buffer,
+    // right-align the samples that we do have (i..e: start with silence).
+    int missingSamples = samples - availableSamples;
+    if (missingSamples > 0) {
+      for (size_t c = 0; c < numChannels; c++) {
+        // Clear the start of the buffer so that we start
+        // the buffer with silence:
+        for (size_t i = 0; i < missingSamples; i++) {
+          outBlock[c][i] = 0.0;
+        }
```
drubinstein, line 71: Please replace this with [std::fill_n](https://en.cppreference.com/w/cpp/algorithm/fill_n)

pedalboard/RubberbandPlugin.h
psobot: Yup - `numChannels` should be very small - 1 or 2, in practice, but potentially up to 8 in the future. We could use a `std::vector`, but doing a `malloc` for what's essentially 16 bytes seems like a bit overkill.

Samples and channels should indeed be constant, so we _could_ allocate this up front.

psobot (approved):

PR #64, opened 2022-01-21
Compensate for plugin latency.

psobot:
Now that we have the `PitchShift` plugin, users of Pedalboard need to manually compensate for the latency added when shifting pitch up or down. This PR fixes that by making Pedalboard latency-aware - automatically compensating when a given plugin returns too few samples in a block.

### DIFF
--- a/pedalboard/ExternalPlugin.h
+++ b/pedalboard/ExternalPlugin.h
@@ -621,8 +621,8 @@ public:
       // only tell Pedalboard to use the last _n_ samples.
       long usableSamplesProduced =
           samplesProvided - pluginInstance->getLatencySamples();
-      return (int)std::min(usableSamplesProduced,
-                           (long)outputBlock.getNumSamples());
+      return static_cast<int>(
+          std::min(usableSamplesProduced, (long)outputBlock.getNumSamples()));
     }
 
     return 0;
--- a/pedalboard/ExternalPlugin.h
+++ b/pedalboard/ExternalPlugin.h
@@ -514,6 +514,7 @@ public:
 
       // Force prepare() to be called again later by invalidating lastSpec:
       lastSpec.maximumBlockSize = 0;
+      samplesProvided = 0;
     }
   }
 
@@ -543,8 +544,8 @@ public:
     }
   }
 
-  void
-  process(const juce::dsp::ProcessContextReplacing<float> &context) override {
+  int process(
+      const juce::dsp::ProcessContextReplacing<float> &context) override {
 
     if (pluginInstance) {
       juce::MidiBuffer emptyMidiBuffer;
@@ -614,7 +615,17 @@ public:
                                            pluginBufferChannelCount,
                                            outputBlock.getNumSamples());
       pluginInstance->processBlock(audioBuffer, emptyMidiBuffer);
+      samplesProvided += outputBlock.getNumSamples();
+
+      // To compensate for any latency added by the plugin,
+      // only tell Pedalboard to use the last _n_ samples.
+      long usableSamplesProduced =
+          samplesProvided - pluginInstance->getLatencySamples();
+      return (int)std::min(usableSamplesProduced,
+                           (long)outputBlock.getNumSamples());
     }
+
+    return 0;
   }
 
   std::vector<juce::AudioProcessorParameter *> getParameters() const {
@@ -634,6 +645,12 @@ public:
     return nullptr;
   }
 
+  virtual int getLatencyHint() override {
+    if (!pluginInstance)
+      return 0;
+    return pluginInstance->getLatencySamples();
+  }
+
 private:
   constexpr static int ExternalLoadSampleRate = 44100,
                        ExternalLoadMaximumBlockSize = 8192;
@@ -641,6 +658,8 @@ private:
   juce::PluginDescription foundPluginDescription;
   juce::AudioPluginFormatManager pluginFormatManager;
   std::unique_ptr<juce::AudioPluginInstance> pluginInstance;
+
+  long samplesProvided = 0;
 
   ExternalPluginReloadType reloadType = ExternalPluginReloadType::Unknown;
 };
--- a/pedalboard/JucePlugin.h
+++ b/pedalboard/JucePlugin.h
@@ -58,12 +58,13 @@ public:
     }
   }
 
-  void process(
-      const juce::dsp::ProcessContextReplacing<float> &context) override final {
+  int process(
+      const juce::dsp::ProcessContextReplacing<float> &context) override {
     dspBlock.process(context);
+    return context.getOutputBlock().getNumSamples();
   }
 
-  void reset() override final { dspBlock.reset(); }
+  void reset() override { dspBlock.reset(); }
 
   DSPType &getDSP() { return dspBlock; };
 
--- a/pedalboard/RubberbandPlugin.h
+++ b/pedalboard/RubberbandPlugin.h
@@ -50,142 +50,86 @@ public:
 
   int process(
       const juce::dsp::ProcessContextReplacing<float> &context) override final {
-    if (!rbPtr) {
-      throw std::runtime_error("Rubber Band plugin failed to instantiate.");
-    }
+    if (rbPtr) {
+      auto inBlock = context.getInputBlock();
+      auto outBlock = context.getOutputBlock();
 
-    auto ioBlock = context.getOutputBlock();
-    auto numChannels = ioBlock.getNumChannels();
+      auto len = inBlock.getNumSamples();
+      auto numChannels = inBlock.getNumChannels();
 
-    if (numChannels > MAX_CHANNEL_COUNT) {
-      throw std::runtime_error(
-          "Pitch shifting or time stretching plugins support a maximum of " +
-          std::to_string(MAX_CHANNEL_COUNT) + " channels.");
-    }
+      jassert(len == outBlock.getNumSamples());
+      jassert(numChannels == outBlock.getNumChannels());
 
-    float *ioChannels[MAX_CHANNEL_COUNT] = {};
-
-    // for (size_t i = 0; i < numChannels; i++) {
-    //   ioChannels[i] = ioBlock.getChannelPointer(i);
-    // }
-
-    // Push all of the input samples we have into RubberBand:
-    // rbPtr->process(ioChannels, ioBlock.getNumSamples(), false);
-    // printf("Pushed %d samples into Rubber Band.\n", ioBlock.getNumSamples());
-    // int availableSamples = rbPtr->available();
-
-    int samplesProcessed = 0;
-    int samplesWritten = 0;
-    while (samplesProcessed < ioBlock.getNumSamples() || rbPtr->available()) {
-      int samplesRequired = rbPtr->getSamplesRequired();
-      int inputChunkLength = std::min(
-          (int)(ioBlock.getNumSamples() - samplesProcessed), samplesRequired);
+      const float **inChannels =
+          (const float **)alloca(numChannels * sizeof(float *));
+      float **outChannels = (float **)alloca(numChannels * sizeof(float *));
 
       for (size_t i = 0; i < numChannels; i++) {
-        ioChannels[i] = ioBlock.getChannelPointer(i) + samplesProcessed;
+        inChannels[i] = inBlock.getChannelPointer(i);
+        outChannels[i] = outBlock.getChannelPointer(i);
       }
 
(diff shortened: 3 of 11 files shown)


pedalboard/RubberbandPlugin.h
```diff
@@ -76,10 +113,23 @@ Base class for rubberband plugins.
...
 
 protected:
+  virtual int getLatencyHint() override {
+    if (!rbPtr)
+      return 0;
+
+    initialSamplesRequired =
+        std::max(initialSamplesRequired,
+                 (int)(rbPtr->getSamplesRequired() + rbPtr->getLatency() +
+                       lastSpec.maximumBlockSize));
```
Johnxjp, line 127: Why add the blockSize as well? The documentation doesn't seem to suggest it's related to latency.

psobot: I'm glad you picked up on this - and I wish I had a good answer here! 😅 I opened [an issue against Rubber Band](https://github.com/breakfastquay/rubberband/issues/54#issuecomment-1018521934) and asked @cannam if there was a strict upper bound on the latency required, and I haven't heard back.

In practice, though, I found that using just the sum of `getSamplesRequired` and `getLatency` gave an underestimate of the latency that would be introduced. Adding `maximumBlockSize` seems to turn that underestimate into a slight overestimate when testing with a variety of different scale factors.

pedalboard/ExternalPlugin.h
```diff
@@ -614,7 +615,17 @@ template <typename ExternalPluginType> class ExternalPlugin : public Plugin {
                                            pluginBufferChannelCount,
                                            outputBlock.getNumSamples());
       pluginInstance->processBlock(audioBuffer, emptyMidiBuffer);
+      samplesProvided += outputBlock.getNumSamples();
```
Johnxjp, line 618: Why is samplesProvided persisted over multiple calls to process?

pedalboard/RubberbandPlugin.h
Johnxjp: Looks like we're conflating the latency and samples required for processing based on Chris' answer. Will it be practical to implement the suggestion of gathering more samples for the first process?

Johnxjp (commented):
Looks good but had a few questions.

pedalboard/RubberbandPlugin.h
psobot: In our case, we do need to consider both values - both the samples required for processing and the latency introduced by processing - as Pedalboard effectively operates on streams of audio, rather than fixed-size buffers. 

Chris mentions that:
> This is all really designed for a pull model in which each processing unit can request as many samples as it needs from its supplier

...but unfortunately, Pedalboard (like the VST API) is designed around a "push" model. Each Pedalboard plugin is pushed a block of samples to process, and is expected to return an identically-sized block. This PR softens that requirement by allowing a plugin to return fewer samples than were provided to it, allowing plugins to buffer audio internally, but a plugin still can't request that the input stream be advanced by a variable amount.

pedalboard/JucePlugin.h
```diff
@@ -58,12 +58,13 @@ template <typename DSPType> class JucePlugin : public Plugin {
     }
   }
 
-  void process(
-      const juce::dsp::ProcessContextReplacing<float> &context) override final {
+  int process(
+      const juce::dsp::ProcessContextReplacing<float> &context) override {
     dspBlock.process(context);
+    return context.getOutputBlock().getNumSamples();
```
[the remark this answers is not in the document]
psobot: None of the internal JUCE plugins that we're using introduce any latency. (Good question, though! I had to go double check. `BlockingConvolution` _could_ introduce latency, but we're using it in a way that guarantees it's latency-free.)

pedalboard/process.h
```diff
@@ -165,75 +262,120 @@ process<float>(const py::array_t<float, py::array::c_style> inputArray,
...
+        pluginSamplesReceived += outputSamples;
+
+        int missingSamples = blockSize - outputSamples;
+        if (missingSamples > 0 && pluginSamplesReceived > 0) {
+          // This can only happen if the plugin we're using is returning us more
+          // than one chunk of audio that's not completely full, which can
+          // happen sometimes. In this case, we would end up with gaps in the
+          // audio output:
+          //               empty  empty  full   part
+          //              [______|______|AAAAAA|__BBBB]
```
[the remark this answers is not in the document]
psobot: That's definitely possible; an improperly-implemented plugin could left-align its audio output if it doesn't have enough samples to fill a buffer. Most third-party plugins are intended for real-time operation, though, in which case a left-aligned output would introduce very clear drop-outs in the audio.

pedalboard/ExternalPlugin.h
psobot: External plugins always have to give us a full buffer (as per the VST/Audio Unit APIs), so we can't ask the plugin how many "useful" samples were actually provided each time we call `processBlock`. Instead, we keep track of the total number of samples provided since this plugin was last reset, and subtract the amount of latency that the plugin told us it adds to the stream. This latency value might be larger than the size of the block provided by a single `process` call, so we need to persist this value between calls to compensate appropriately.

pedalboard/ExternalPlugin.h
```diff
@@ -543,8 +544,8 @@ template <typename ExternalPluginType> class ExternalPlugin : public Plugin {
     }
   }
 
-  void
-  process(const juce::dsp::ProcessContextReplacing<float> &context) override {
+  int process(
```
drubinstein, line 547: Any thoughts on making this an [int64_t](https://en.cppreference.com/w/cpp/types/integer) or maybe an unsigned long? Looks like unsigned at the min as negative return shouldn't be possible

pedalboard/ExternalPlugin.h
```diff
@@ -614,7 +615,17 @@ template <typename ExternalPluginType> class ExternalPlugin : public Plugin {
                                            pluginBufferChannelCount,
                                            outputBlock.getNumSamples());
       pluginInstance->processBlock(audioBuffer, emptyMidiBuffer);
+      samplesProvided += outputBlock.getNumSamples();
+
+      // To compensate for any latency added by the plugin,
+      // only tell Pedalboard to use the last _n_ samples.
+      long usableSamplesProduced =
+          samplesProvided - pluginInstance->getLatencySamples();
+      return (int)std::min(usableSamplesProduced,
```
drubinstein, line 624: Optional - use static_cast instead because even though it's more characters it's more "correct"

pedalboard/ExternalPlugin.h
```diff
@@ -642,6 +659,8 @@ template <typename ExternalPluginType> class ExternalPlugin : public Plugin {
   juce::AudioPluginFormatManager pluginFormatManager;
   std::unique_ptr<juce::AudioPluginInstance> pluginInstance;
 
+  long samplesProvided = 0;
```
drubinstein, line 662: Similar comment about using unsigned here.

pedalboard/plugins/PitchShift.h
```diff
@@ -23,41 +23,38 @@ namespace py = pybind11;
...
 
+  // Allow pitch shifting by up to 6 octaves up or down:
+  static constexpr float MIN_SCALE_FACTOR = 0.015625;
+  static constexpr float MAX_SCALE_FACTOR = 64.0;
+
 public:
   void setScaleFactor(double scale) {
-    if (scale <= 0) {
-      throw std::range_error("Pitch scale must be a value greater than 0.0.");
+    if (scale < MIN_SCALE_FACTOR || scale > MAX_SCALE_FACTOR) {
```
drubinstein, line 41: For c++20 nerds - looks like what you want is a [spaceship](https://en.cppreference.com/w/cpp/language/operator_comparison#Three-way_comparison)

pedalboard/process.h
```diff
@@ -57,40 +57,51 @@ processSingle(const py::array_t<SampleType, py::array::c_style> inputArray,
...
-               double sampleRate, const std::vector<Plugin *> &plugins,
-               unsigned int bufferSize, bool reset) {
+template <typename T>
+ChannelLayout
+detectChannelLayout(const py::array_t<T, py::array::c_style> inputArray) {
+  py::buffer_info inputInfo = inputArray.request();
+
+  if (inputInfo.ndim == 1) {
+    return ChannelLayout::Interleaved;
+  } else if (inputInfo.ndim == 2) {
```
drubinstein, line 67: Maybe add some log here about inferring a shape. May help in future debugging. I would consider having a default layout and let the user override the layout instead.

pedalboard/process.h
```diff
@@ -102,19 +113,105 @@ process<float>(const py::array_t<float, py::array::c_style> inputArray,
...
+  // TODO: Avoid the need to copy here if offsetSamples is 0!
+  py::array_t<T> outputArray;
+  if (ndim == 2) {
+    switch (channelLayout) {
+    case ChannelLayout::Interleaved:
+      outputArray = py::array_t<T>({outputSampleCount, numChannels});
+      break;
+    case ChannelLayout::NotInterleaved:
+      outputArray = py::array_t<T>({numChannels, outputSampleCount});
+      break;
```
drubinstein, line 160: Usually it's a good idea to have a default here with an error throw just in case. Shouldnt be possible theoretically, but you never know.

pedalboard/process.h
```diff
@@ -102,19 +113,105 @@ process<float>(const py::array_t<float, py::array::c_style> inputArray,
...
-                          : py::array_t<float>(numSamples);
   py::buffer_info outputInfo = outputArray.request();
 
+  // Depending on the input channel layout, we need to copy data
+  // differently. This loop is duplicated here to move the if statement
+  // outside of the tight loop, as we don't need to re-check that the input
+  // channel is still the same on every iteration of the loop.
+  T *outputBasePointer = static_cast<T *>(outputInfo.ptr);
+
+  switch (channelLayout) {
```
drubinstein, line 174: With this number of switch cases I'd almost consider in a future pr abstracting these functions into some interface or in the over engineered C++17 case, use [std::variant](https://en.cppreference.com/w/cpp/utility/variant)

pedalboard/process.h
```diff
@@ -165,75 +262,120 @@ process<float>(const py::array_t<float, py::array::c_style> inputArray,
...
+            // Only move the samples received before this latest block was
+            // rendered, as audio is right-aligned within blocks by convention.
+            int samplesToMove = pluginSamplesReceived - outputSamples;
+            float *outputStart =
+                ioBuffer.getWritePointer(c) + totalOutputLatencySamples;
+            float *expectedOutputEnd =
+                ioBuffer.getWritePointer(c) + blockEnd - outputSamples;
+            float *expectedOutputStart = expectedOutputEnd - samplesToMove;
+
+            std::memmove((char *)expectedOutputStart, (char *)outputStart,
```
drubinstein, line 340: Consider [std::reverse_copy](https://en.cppreference.com/w/cpp/algorithm/reverse_copy) or [std::copy_n](https://en.cppreference.com/w/cpp/algorithm/copy_n) with a reverser iterator. std::memmove can work with overlaps, but i think copying in reverse is safer. May also be more efficient too, but I can't confirm that. For all I know memmove already acts that way.

pedalboard/ExternalPlugin.h
psobot: That's true - I'm a bit split on how to handle that, though. If we make this an unsigned int, it's really easy to subtract from the result by accident and underflow. I agree we should definitely make this something larger than `int`, though - an `int` can only count up to about 13.5 hours worth of samples at 44.1k, which is something that could be feasibly exceeded in some scenarios.

pedalboard/process.h
psobot: `std::memmove` [seems to guarantee this](https://en.cppreference.com/w/cpp/string/byte/memmove):
> Despite being specified "as if" a temporary buffer is used, actual implementations of this function do not incur the overhead of double copying or extra memory. For small count, it may load up and write out registers; for larger blocks, a common approach (glibc and bsd libc) is to copy bytes forwards from the beginning of the buffer if the destination starts before the source, and backwards from the end otherwise, with a fall back to `std::memcpy` when there is no overlap at all.

pedalboard/ExternalPlugin.h
psobot: Actually, on further thought: this will only ever happen if `buffer_size` is greater than 13.5 hours, or 2<sup>31</sup> - 1 samples, or 8GB. That's large, but not out of the realm of possibility, especially if working at higher sample rates - I'll make a separate PR to handle absurdly large audio files or buffer lengths.

PR #65, opened 2022-01-24
Change PitchShift plugin interface to use semitones.

psobot:
Slight API cleanup: rather than exposing the raw scale factor from Rubber Band, we might want to expose a more intuitive value like the number of semitones being shifted by. This change assumes all users of this plugin want [12-tone equal temperament](https://en.wikipedia.org/wiki/Equal_temperament), but I think that's a safe assumption to make for now.

### DIFF
--- a/pedalboard/plugins/PitchShift.h
+++ b/pedalboard/plugins/PitchShift.h
@@ -35,7 +35,7 @@ private:
   // Allow pitch shifting by up to 6 octaves up or down:
   static constexpr int MIN_SEMITONES = -6 * 12;
   static constexpr int MAX_SEMITONES = 6 * 12;
-  
+
   double getScaleFactor() { return pow(2, (getSemitones() / 12)); }
 
 public:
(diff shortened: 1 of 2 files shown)


pedalboard/plugins/PitchShift.h
```diff
@@ -30,48 +30,45 @@ class PitchShift : public RubberbandPlugin
...
+    _semitones = semitones;
 
     if (rbPtr)
-      rbPtr->setPitchScale(_scaleFactor);
+      rbPtr->setPitchScale(getScaleFactor());
   }
 
-  double getScaleFactor() { return _scaleFactor; }
+  double getSemitones() { return _semitones; }
+  double getScaleFactor() { return pow(2, (getSemitones() / 12)); }
```
Johnxjp, line 54: Do you want to expose an option to retrieve scale factor in the python bindings as well? Otherwise is it worth making this function private / protected?

psobot: Good idea - I think we might want to keep the scale factor private (to avoid complicating the Python API) so I'll make this private for clarity.

PR #67, opened 2022-01-26
Add VST3 preset loading API.

emilio1234:
Contributed: Loading VST3 preset files (.vstprest). Test for preset loading.


pedalboard/ExternalPlugin.h
```diff
@@ -226,6 +226,31 @@ template <typename ExternalPluginType> class ExternalPlugin : public Plugin {
...
+      void visitVST3Client (const juce::ExtensionsVisitor::VST3Client& client) override{
+          result = false;
+          juce::File fl(presetFile);
+          juce::MemoryBlock mb;
+
+          if (fl.loadFileAsData(mb)){
+              result =  client.setPreset(mb);
+          }
+      }
+  };
```
psobot, line 244: Rather than quietly returning `false` if a plugin fails to load, [the Pythonic way](https://docs.python.org/2/glossary.html#term-eafp) might be to throw an exception. Here's a suggestion that throws two different exceptions for different errors.

```suggestion
  struct PresetVisitor : public juce::ExtensionsVisitor {
      const std::string presetFilePath;
      
      PresetVisitor(const std::string presetFilePath): presetFilePath(presetFilePath) { }
      
      void visitVST3Client(const juce::ExtensionsVisitor::VST3Client& client) override {
          juce::File presetFile(presetFilePath);
          juce::MemoryBlock presetData;

          if (!presetFile.loadFileAsData(presetData)) {
            throw std::runtime_error("Failed to read preset file: " + presetFilePath);
          }

          if (!client.setPreset(presetData)) {
            throw std::runtime_error("Plugin returned an error when loading data from preset file: " + presetFilePath);
          }
      }
  };
```

pedalboard/ExternalPlugin.h
```diff
@@ -226,6 +226,31 @@ template <typename ExternalPluginType> class ExternalPlugin : public Plugin {
...
+
+
+  bool loadPresetData(std::string stateFile) {
+      PresetVisitor visitor {stateFile};
+      pluginInstance->getExtensions(visitor);
+      return visitor.result;
+  }
```
psobot, line 251: Together with the above suggestion to raise an exception:

```suggestion
  void loadPresetData(std::string presetFilePath) {
      PresetVisitor visitor {presetFilePath};
      pluginInstance->getExtensions(visitor);
  }
```

pedalboard/ExternalPlugin.h
```diff
@@ -793,6 +818,10 @@ inline void init_external_plugins(py::module &m) {
              ss << ">";
              return ss.str();
            })
+      .def("load_preset",
+          &ExternalPlugin<juce::VST3PluginFormat>::loadPresetData,
+          "Load a VST3 preset file .vstprest format "
+          "return false if fails")
```
psobot, line 824: It might be useful to include the argument name here, so that people can tell what the argument is called in Python (rather than the default, which I think is `arg0`).

```suggestion
      .def("load_preset",
          &ExternalPlugin<juce::VST3PluginFormat>::loadPresetData,
          "Load a VST3 preset file in .vstpreset format.", py::arg("preset_file_path"))
```

psobot (approved):
Thanks @emilio1234! This looks really great. I've made a couple of suggestions for how the code could be improved in-line - feel free to use GitHub's `Commit suggestion` functionality if you want to quickly apply the suggestions.

pedalboard/ExternalPlugin.h
emilio1234: agree

emilio1234:
I've tested locally the changes. The function throws exceptions on error. Looks good by me.

PR #69, opened 2022-01-26
Fix audio glitches when running VST3 or AU plugins whose reported latency exceeds buffer size.

psobot:
Title says it all - a subtle bug slipped through thanks to negative numbers being returned from a function that should only ever return positive numbers. Thanks @drubinstein for the inspiration to check this.


drubinstein (commented):
Wish there was a test e.g. with a "bad" plugin, but lgtm

drubinstein (approved):

PR #71, opened 2022-01-26
Add MP3Compressor plugin.

psobot:
This PR adds `MP3Compressor`, a basic wrapper around [LAME](https://lame.sourceforge.io/), the "canonical" open-source, GPL-compatible MP3 compressor that's been around for ~24 years. This plugin compresses and then decompresses the audio stream with LAME.

(Note that the resulting output from Pedalboard is still in the same uncompressed format; this plugin is intended to apply the _sound_ of compression, rather than returning an output that is itself in a compressed format.)

This is still a draft PR as there are a couple things remaining to clean up:
 - `setup.py` has been refactored, and the changes added (to support building the LAME C code) have not yet been tested on Windows
 -  Mono audio is not yet supported (or rather, not tested)
 - Tests are failing for `-V0` and `-V0.5` compression levels (the highest possible)
 - Arbitrary sample rates have not been tested, and will likely throw exceptions (but these should be listed in the tests)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -98,6 +98,7 @@ else:
     LAME_FLAGS.append(f"-include{LAME_CONFIG_FILE}")
 ALL_CFLAGS.extend(LAME_FLAGS)
 ALL_SOURCE_PATHS += list(Path("vendors/lame/libmp3lame").glob("*.c"))
+ALL_SOURCE_PATHS += list(Path("vendors/lame/libmp3lame/vector").glob("*.c"))
 ALL_SOURCE_PATHS += list(Path("vendors/lame/mpglib").glob("*.c"))
 ALL_INCLUDES += [
     'vendors/lame/include/',
--- a/setup.py
+++ b/setup.py
@@ -18,99 +18,151 @@
 import os
 import platform
 from subprocess import check_output
-from pybind11.setup_helpers import Pybind11Extension
+from pybind11.setup_helpers import Pybind11Extension, build_ext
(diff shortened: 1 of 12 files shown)


setup.py
```diff
@@ -107,8 +107,11 @@
         "-DPROTOTYPES=1",
         "-DHAVE_STRCHR",
         "-DHAVE_MEMCPY",
-        "-Dieee754_float32_t=float",
         "-DHAVE_MPGLIB",
+        # Some data types are defined in config.h, which we don't include:
+        "-Dieee754_float32_t=float",
+        "-Duint32_t=u_int32_t",
+        "-Duint16_t=u_int16_t",
```
drubinstein, line 114: This feels a little dangerous. [cstdint](https://en.cppreference.com/w/cpp/types/integer) already defines a lot of these so wouldnt you risk overwriting those defines? What config.h are you referring to?

psobot: You're right, this shouldn't be inserted when we compile C++. LAME is unfortunately all C, and part of its build process involves [`#define`-ing a bunch of stuff](https://github.com/lameproject/lame/blob/master/config.h.in) in a `config.h`.

We might have to find another way to either generate this `config.h` on-the-fly during compilation, or to inject the same definitions that it does so that the C code compiles. (Another option might be to check in a copy of `config.h` in `vendors`, and then [use the `-include` flag](https://stackoverflow.com/a/3387518/679081) to force its inclusion when compiling C.)

PR #72, opened 2022-01-30
Add GSMFullRateCompressor plugin.

psobot:
This PR does what it says on the tin: adds a [GSM, or "Full Rate"](https://en.wikipedia.org/wiki/Full_Rate) plugin, to make audio sound like it was transmitted over a 2G cell phone. To do this, this PR compiles in `libgsm`, a free, open-source, and ISC-licensed GSM encoder/decoder plugin.


drubinstein (commented):
I'm gonna be a very pedantic, but I wouldn't call this a `GSMCompressor` but a `GSMFullRateCompressor` (though I'd prefer `GSMFullRateEncoder`) or allow this plugin to take in an enum which gives the appropriate codec parameters.. GSM supports multiple audio bitrates including [half rate](https://en.wikipedia.org/wiki/Half_Rate), [enhanced full rate](https://en.wikipedia.org/wiki/Enhanced_full_rate) and [adaptive multi rate](https://en.wikipedia.org/wiki/Adaptive_Multi-Rate_audio_codec). HR and FR audio sounds incredibly different from each other.

drubinstein:
Then maybe not call it either and label the plugin `GSMFullRate` ?

PR #74, opened 2022-02-03
Fix boundary effects before (and after) some pitch shift operations.

psobot:
This PR fixes a small issue with the `PitchShift` plugin: the first couple hundred samples introduce a slight distortion to the audio, even if setting the number of semitones of shift to `0.0`.

To avoid this distortion, this PR adds a fixed amount (1 second) of silence before beginning rendering, which seems to prime the internal buffers of the Rubber Band library. Rather than adding this silence just to the `PitchShift` plugin, this PR adds a generic `PrimeWithSilence<T>` plugin class, which can be used to wrap an existing plugin and pass a specific amount of silence before rendering begins, while automatically compensating for latency.

Before this PR, testing with a sine wave input:
image

And after this PR:
image

### DIFF
--- a/pedalboard/plugin_templates/PrimeWithSilence.h
+++ b/pedalboard/plugin_templates/PrimeWithSilence.h
@@ -1,0 +1,179 @@
+/*
+ * pedalboard
+ * Copyright 2022 Spotify AB
+ *
+ * Licensed under the GNU Public License, Version 3.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    https://www.gnu.org/licenses/gpl-3.0.html
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include "../JuceHeader.h"
+#include "../Plugin.h"
+#include "../plugins/AddLatency.h"
+#include <mutex>
+
+namespace Pedalboard {
+
+/**
+ * A dummy plugin that buffers audio data internally, used to test Pedalboard's
+ * automatic latency compensation.
+ */
+template <typename T, typename SampleType = float>
+class PrimeWithSilence
+    : public JucePlugin<juce::dsp::DelayLine<
+          SampleType, juce::dsp::DelayLineInterpolationTypes::None>> {
+public:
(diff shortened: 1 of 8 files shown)


pedalboard/plugin_templates/PrimeWithSilence.h
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  T &getNestedPlugin() { return plugin; }
+
+  void setSilenceLengthSamples(int newSilenceLengthSamples) {
+    this->getDSP().setMaximumDelayInSamples(newSilenceLengthSamples);
+    this->getDSP().setDelay(newSilenceLengthSamples);
+    silenceLengthSamples = newSilenceLengthSamples;
+
+    reset();
```
Johnxjp, line 77: Why is the reset necessary here? Wonder if should remove this set function in general and just force the creation of a new object?

tests/test_pitch_shift.py
```diff
@@ -50,15 +50,13 @@ def test_pitch_shift_extremes(semitones, sample_rate, buffer_size):
...
+def test_pitch_shift_latency_compensation(semitones, fundamental_hz, sample_rate, buffer_size):
+    num_seconds = 5.0
     samples = np.arange(num_seconds * sample_rate)
     sine_wave = np.sin(2 * np.pi * fundamental_hz * samples / sample_rate)
     plugin = Pedalboard([PitchShift(semitones), PitchShift(-semitones)])
     output = plugin.process(sine_wave, sample_rate, buffer_size=buffer_size)
-    np.testing.assert_allclose(
-        sine_wave[sample_rate:-sample_rate], output[sample_rate:-sample_rate], rtol=0.01, atol=0.01
-    )
+    np.testing.assert_allclose(sine_wave, output, rtol=0.01, atol=0.01)
```
Johnxjp, line 62: Is it worth adding a stricter tolerance now?

Johnxjp (commented):
Is there a way we can solve this problem by just augmenting the pitch shift plugin? This issue for now is localised to Rubberband and creating a new generic plugin could be premature. What do you think?

tests/test_pitch_shift.py
psobot: Yup! Looks like we can lower this `atol` value from `0.01` to `1e-06` (10,000x stricter) and this test still passes.

psobot (commented):
Thanks @Johnxjp!

> This issue for now is localised to Rubberband and creating a new generic plugin could be premature. What do you think?

That's true at the moment, but there are already a couple of plugins that would benefit from using this `PrimeWithSilence` wrapper to simplify their logic (`MP3Compressor` and the upcoming [`GSMFullRateCompressor`](https://github.com/spotify/pedalboard/pull/72), for example).

PR #85, opened 2022-03-07
Add pedalboard.io.AudioFile.

psobot:
This PR adds:
 - `pedalboard.io`, a new submodule containing audio input/output classes
 - `pedalboard.io.AudioFile`, a base class for interfacing with audio files
 - `pedalboard.io.ReadableAudioFile`, a concrete class for reading audio files
 - `pedalboard.io.WriteableAudioFile`, a concrete class for writing audio files
 - Tests for the above
 - Examples in the README.

Questions that may arise from this change:
 - _Why are these useful or necessary?_
   - Most Python audio reading/writing libraries are single-format ([`pymad`](https://github.com/jaqx0r/pymad), [`pyogg`](https://github.com/TeamPyOgg/PyOgg), etc), require DLLs to be installed ([`PySoundFile`](https://pysoundfile.readthedocs.io/en/latest/)), or sometimes shell out to FFMPEG ([`audioread`](https://github.com/beetbox/audioread), [`librosa`](https://librosa.org/doc/main/generated/librosa.load.html)). Pedalboard (by virtue of JUCE) has the ability to provide high-performance audio reading and writing functionality out-of-the-box without much code, removing the need for third-party dependencies. In tests, this code is up to 20x faster than `librosa.load` in certain cases (without resampling).
 - _Should this be part of Pedalboard, or a separate library?_
   - Given that Pedalboard already includes JUCE under the hood, and that adding these classes inflates the binary size by only kilobytes, it seems pragmatic to include these classes here rather than creating a separate library that would itself be a couple MB in size.
 - _Why are not all formats supported?_
   - This PR wraps JUCE's `AudioFormatReader` and `AudioFormatWriter` interfaces, which only support reading and writing a certain set of audio formats out-of-the-box. Notably, Pedalboard already includes `libmp3lame`, which could be used to create a custom `AudioFormatWriter` subclass to allow for MP3 file writing out-of-the-box.


faroit:
@psobot nice PR! 

* did you test seeking performance against ...e.g. soundfile?
* would you be able to create PR for https://github.com/faroit/python_audio_loading_benchmark/ ?

keunwoochoi:
this is great!

PR #95, opened 2022-03-23
Re-enable x86-only wheels on macOS.

### LINKED ISSUE #94: Can't install version higher than 0.3.8 on macos rosetta
I'm using python 3.8.12 packaged by conda-forge on an M1 and when I try to pip install pedalboard, the latest version I get is 0.3.8. If I try to manually download wheels and install those it tells me my platform is unsupported. The python version I'm using is in Rosetta.

psobot:
Fixes #94.

When running Python on Apple Silicon under [Rosetta 2](https://support.apple.com/en-us/HT211861), Pip looks for only `macosx_x86_64` wheels, not `universal2` wheels. This PR makes debugging easier by re-publishing x86_64 wheels on macOS, as we previously did until version 0.3.8.

### DIFF
--- a/.github/workflows/all.yml
+++ b/.github/workflows/all.yml
@@ -177,6 +177,9 @@ jobs:
         include:
         - { os: macos-latest, build: cp36-macosx_x86_64 }
         - { os: macos-latest, build: cp37-macosx_x86_64 }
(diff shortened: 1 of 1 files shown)


lminer:
Looks like the version didn't get bumped in the filename.

PR #98, opened 2022-04-14
Add experimental live audio support.

psobot:
Still very very WIP.

TODO:
 - [x] Add proper locking and/or lock-free synchronization to ensure that the audio thread always has a valid view of the plugins used for processing
 - [x] Add checks to ensure we don't route built-in microphone directly back out to built-in speakers (causing feedback)
 - [x] Add a better API for getting device names and input/output details
 - [x] Simplify the `stream.plugins` API - it should be a "real" `list` object that supports `.append(Plugin)`, etc.
 - [x] Should the `AudioStream` API block while audio is being streamed? Would avoid creating a stream, processing audio, and then having it "get lost"
 - [x] Test what happens if an `AudioStream` gets deallocated while the stream is active
 - [x] Tests in general
 - [x] Verify binary size increase is only ~kilobytes (seems to be ~50kB on macOS)


stan-kirdey:
really interesting work, looking forward to more experiments in live audio processing!

ambv:
I'm just popping in here to say I'm waiting for this! In particular I'd like to replace my use of [pyminiaudio](https://github.com/irmen/pyminiaudio) which doesn't properly handle audio input in my case.

PR #101, opened 2022-04-25
Avoid exposing internal symbols to decrease binary size.

psobot:
This PR turns on [Clang's ThinLTO](https://clang.llvm.org/docs/ThinLTO.html), which should reduce binary size by at least 33% on macOS alone.


setup.py
```diff
@@ -117,8 +117,14 @@
 if platform.system() == "Darwin":
```
carloserocha, line 117: Hi psobot, how are you? I'm ashamed and I could be wrong but,

What you think about decouple the arrays append in function by the OS?

PR #107, opened 2022-06-09
Fix changing plugin parameters example code in README.md

nashaad:
The example code under `Does Pedalboard support changing a plugin's parameters` in README.md is incomplete and didn't run in current state. Additionally, when trying to modify it to work, I noticed that it wasn't iterating through the correct ndarray needed to show the percentage which the code was also trying to do.

Completed the input, output parts, changes to the chunking code, in keeping with the simplicity of the original (i.e. without using `sounddevice` library). Added tqdm to show a progress bar, and verified by testing with sample audio.

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,29 +245,45 @@ Yes! While there's no built-in function for this, it is possible to
 vary the parameters of a plugin over time manually:
 
 ```python
-import numpy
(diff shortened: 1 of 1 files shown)


nashaad:
@psobot looks good :) processing both in and out in chunks makes sense

PR #110, opened 2022-06-23
Midi support

robclouth:
Summary: I've started a basic implementation of MIDI support to get things started

Problem
Pedalboard doesn't currently support plugin instruments because there is no way of sending MIDI to the plugins

Solution
I've added very basic MIDI support using juce::MidiMessageSequence and juce::MidiBuffer.

Result
This is mostly to get the ball rolling with MIDI support, it's definitely not ready for prime-time and haven't rigorously followed the contribution guide. It works for single processors, but not chains because the note off doesn't seem to work for chains. I haven't written any tests because I'm reckless. It's got to where I need it so I'll give it a rest for while. If someone wants to pick up the baton feel free!

For now I've done the following:

- Passed juce::MidiMessageSequence to the top-level process methods and based on the block time, passed MidiBuffers to the plugin process methods. This causes some complications with the Chains. My ugly non-working solution is to overload the process method and convert the MidiBuffer back into MidiMessageSequence
- Added note on and off generation with mido
- Added a second parameter to the plugin.process and pedalboard process methods that takes midi notes with time offsets in seconds. This should be optional but I couldn't quickly figure out how to do that. If passing in MIDI then the input audio should be optional too I guess.

E.g.

```
from pedalboard import Pedalboard, load_plugin
from pedalboard.io import AudioFile
from pedalboard.midi import note_on, note_off

import numpy as np

# Load a VST3 or Audio Unit plugin from a known path on disk:
vst = load_plugin("TAL-NoiseMaker.vst3")

print(vst.parameters.keys())

output_audio = np.zeros((44100, 2))
board = Pedalboard([vst])

effected = board(output_audio, [note_on(0.0, 72, 127), note_off(0.2, 72, 0)], 44100)

with AudioFile('processed-output.wav', 'w', 44100, effected.shape[1]) as f:
  f.write(effected)
```


pedalboard/python_bindings.cpp
```diff
@@ -71,28 +71,28 @@ PYBIND11_MODULE(pedalboard_native, m) {
   m.def("process", process<float>,
         "Run a 32-bit floating point audio buffer through a list of Pedalboard "
         "plugins.",
-        py::arg("input_array"), py::arg("sample_rate"), py::arg("plugins"),
+        py::arg("input_array"), py::arg("midi_messages"), py::arg("sample_rate"), py::arg("plugins"),
```
psobot, line 74: Unfortunately I don't think we can change this API like this; lots of user code passes `sample_rate` as the second argument to `process` (or `__call__`) without specifying its keyword argument name. If we merge this, all of that user code will break as it will interpret the second argument as `midi_messages`.

This change also makes it mandatory to pass MIDI messages every time a plugin is invoked. If we move this to the last argument, we can also make it optional and provide `None` or `[]` as a default value, to avoid adding complexity for those who don't use this feature.

requirements.txt
```diff
@@ -1 +1,2 @@
-numpy
\ No newline at end of file
+numpy
+midi==1.2.10
```
psobot, line 2: Two things:
 - I don't think this import is actually used (we seem to `import mido` instead of `import midi` elsewhere)
 - I'm pretty hesitant to add another runtime dependency if we can avoid it; MIDI messages are fairly simple to represent (as a five-tuple of `(type, note, channel, velocity, time)`) and we could use [duck typing](https://en.wikipedia.org/wiki/Duck_typing) to accept objects from the `mido` package (or the `midi` package) without adding them as a runtime dependency.

pedalboard/MidiUtils.h
```diff
@@ -0,0 +1,53 @@
...
+#pragma once
+#include "JuceHeader.h"
+
+#include <pybind11/numpy.h>
+#include <pybind11/pybind11.h>
+
+namespace Pedalboard {
+
+juce::MidiMessageSequence
+copyPyArrayIntoJuceMidiMessageSequence(const py::array_t<float, py::array::c_style> midiMessages) {
```
psobot, line 27: This helper seems to re-interpret floating-point data into integer data, which seems a bit risky and confusing. Given that MIDI messages are probably going to be sparse (i.e.: not nearly as large as audio data) we could probably just use a `std::vector<std::tuple<int, int, int, float>>` or could create our own thin wrapper class and use that (i.e.: `std::vector<pedalboard.MIDIMessage>`).

pedalboard/MidiUtils.h
```diff
@@ -0,0 +1,53 @@
...
+
+  for(int i = 0; i < inputInfo.shape[0]; i++) {
+    int byte1 = (int) data[i * 4 + 0];
+    int byte2 = (int) data[i * 4 + 1];
+    int byte3 = (int) data[i * 4 + 2];
+    float timeSeconds = data[i * 4 + 3];
+
+    midiSequence.addEvent(juce::MidiMessage(byte1, byte2, byte3), timeSeconds);
+
+    DBG( byte1 );
```
psobot, line 47:
```suggestion
```

pedalboard/ExternalPlugin.h
```diff
@@ -753,10 +753,10 @@ template <typename ExternalPluginType> class ExternalPlugin : public Plugin {
   }
 
   int process(
-      const juce::dsp::ProcessContextReplacing<float> &context) override {
+      const juce::dsp::ProcessContextReplacing<float> &context, juce::MidiBuffer &midiBuffer) override {
 
     if (pluginInstance) {
-      juce::MidiBuffer emptyMidiBuffer;
+      ;
```
psobot, line 759:
```suggestion
```

psobot (changes requested):

rhelsing:
@psobot Hey there, whats the current status on this PR?

robclouth:
Hey. I won't be continuing on this PR. I got it to where I needed it and put it here in case anyone else found it useful. If anyone wants to continue feel free.

TylerMclaughlin:
hi @psobot et al.

i'm trying to get @robclouth's  MIDI branch working with the example rob posted.  

Seems like a holy grail for people like me who want to render midi to audio faster than real time. ( I'm not a pybind programmer so don't get ur hopes up.  )

Sadly,  i'm getting an error on this branch at the load_plugin phase (_upstream_ of the MIDI input).

`'ValueError: Plugin 'Vital' does not accept audio input. It may be an instrument plug-in and not an audio effect processor.`

This... would make sense given Vital is a VST MIDI instrument, not an audio effect VST.  But I also get the same error for loading a  reverb:
```python
load_plugin("/Library/Audio/Plug-Ins/VST3/ValhallaShimmer.vst3")
```
The issue is similar  to #118 , but I'm on Mac OS 10.15, not Windows.

Anyway, these are the steps I used to build Rob's branch on my machine.  Maybe I'm doing something wrong??

```bash
#basic build, but with the midi branch
git clone --recurse-submodules --shallow-submodules https://github.com/robclouth/pedalboard.git
cd pedalboard
pip3 install pybind11
pip3 install .
# required for this implementation
pip3 install mido
# not sure if needed
git fetch
# necessary to get to the branch with midi support
git switch midi-support
```
`git log` confirms i'm at rob's latest 'note on' commit, with SHA 2559a400cc3dca57fb28536cade3ea75429bd620.

any ideas or workarounds?

PR #117, opened 2022-07-12
Add type hinting and stub files for editor autocomplete.

psobot:
This PR should add type hints to allow editors like VS Code and PyCharm to autocomplete classes, functions, and constants in Pedalboard. A test is also added to ensure that stub files don't drift from the actual runtime implementation.

TODO: how can we automatically copy documentation from the runtime `__doc__` attribute(s) into the stub files as well?


pedalboard/__init__.pyi
```diff
@@ -0,0 +1,297 @@
...
+
+class HighpassFilter(Plugin):
+    cutoff_frequency_hz: float
+    def __init__(self, cutoff_frequency_hz: float = ...) -> None: ...
+
+class Invert(Plugin):
+    def __init__(self) -> None: ...
+
+class LadderFilter(Plugin):
+    class Mode:
```
drubinstein, line 88: I think this can be retyped as an Enum? Hard to say as I dont know how well that will translate.

The comments I left in this class also apply to all the other c++ enums in the PR e.g. in resample

pedalboard/__init__.pyi
```diff
@@ -0,0 +1,297 @@
...
+# that leads you here, feel free to modify the file manually as well.
+#
+# This file may diverge from the runtime/native implementation of the code,
+# as long as the interface remains the same. (i.e.: pure-Python classes
+# may be listed here even if they override/inherit from native classes.)
+
+from typing import Any, ClassVar, Dict, Iterator, List, Optional, Type, Union, overload
+
+import numpy
+from numpy.typing import NDArray
```
drubinstein, line 12: Just a note, this may not be compatible with older versions of numpy/python. Numpy is currently unpinned so it's probably fine, but existing with TF<=2.7 may be an issue for example.

I think to include this you'll have to pin numpy in the repo to >=1.21 (maybe 1.20 I can't remember)

pedalboard/__init__.pyi
```diff
@@ -0,0 +1,297 @@
...
+class HighpassFilter(Plugin):
+    cutoff_frequency_hz: float
+    def __init__(self, cutoff_frequency_hz: float = ...) -> None: ...
+
+class Invert(Plugin):
+    def __init__(self) -> None: ...
+
+class LadderFilter(Plugin):
+    class Mode:
+        __members__: ClassVar[dict] = ...  # read-only
```
drubinstein, line 89: Should be `Dict[]`. I dont believe `dict` is understood as a type until py3.9 or so. This is a read only class so 🤷

pedalboard/__init__.pyi
```diff
@@ -0,0 +1,297 @@
...
+        BPF12: ClassVar[LadderFilter.Mode] = ...
+        BPF24: ClassVar[LadderFilter.Mode] = ...
+        HPF12: ClassVar[LadderFilter.Mode] = ...
+        HPF24: ClassVar[LadderFilter.Mode] = ...
+        LPF12: ClassVar[LadderFilter.Mode] = ...
+        LPF24: ClassVar[LadderFilter.Mode] = ...
+        __entries: ClassVar[dict] = ...
```
drubinstein, line 96: Same comment on `dict` vs `Dict`

pedalboard/io/__init__.pyi
```diff
@@ -0,0 +1,112 @@
...
+class AudioFile(object):
+    @overload
+    def __new__(cls, filename: str, mode: Literal["r"] = ...) -> ReadableAudioFile: ...
+    @overload
+    def __new__(cls, file_like: BinaryIO, mode: Literal["r"] = ...) -> ReadableAudioFile: ...
+    @overload
+    def __new__(
+        cls,
+        filename: str,
+        mode: Literal["w"],
```
drubinstein, line 94: I believe Literals `"w", "wb", "r", "rb", "a"` would all be valid here too depending on the situation.

pedalboard/io/__init__.pyi
```diff
@@ -0,0 +1,112 @@
...
+    ) -> WriteableAudioFile: ...
+    @overload
+    def __new__(
+        cls,
+        file_like: BinaryIO,
+        mode: Literal["w"],
+        samplerate: Optional[float] = ...,
+        num_channels: int = ...,
+        bit_depth: int = ...,
+        quality: Union[None, str, float] = ...,
```
drubinstein, line 108:
```suggestion
        quality: Optional[str, float] = ...,
```

pedalboard/io/__init__.pyi
```diff
@@ -0,0 +1,112 @@
...
+    def __new__(cls, file_like: BinaryIO, mode: Literal["r"] = ...) -> ReadableAudioFile: ...
+    @overload
+    def __new__(
+        cls,
+        filename: str,
+        mode: Literal["w"],
+        samplerate: Optional[float] = ...,
+        num_channels: int = ...,
+        bit_depth: int = ...,
+        quality: Union[None, str, float] = ...,
```
drubinstein, line 98:
```suggestion
        quality: Optional[str, float] = ...,
```

pedalboard/pedalboard.py
```diff
@@ -197,8 +203,8 @@ def __init__(self, plugin, parameter_name, search_steps: int = 1000):
         self.ranges: Dict[Tuple[float, float], Union[str, float, bool]] = {}
 
         with self.__get_cpp_parameter() as cpp_parameter:
-            start_of_range = 0
-            text_value = None
+            start_of_range: float = 0
+            text_value: Union[None, str] = None
```
drubinstein, line 207:
```suggestion
            text_value: Optional[str] = None
```

drubinstein (approved):
cool

pedalboard/io/__init__.pyi
psobot: Is this a new feature in newer Pythons? I get:

```python
In [3]: Optional[str, float]
...

TypeError: Optional[t] requires a single type. Got (<class 'str'>, <class 'float'>).
```

pedalboard/io/__init__.pyi
drubinstein: Curious, if the mode can only take one value, why include mode as an argument at all (other than for backwards compatibility now)

pedalboard/io/__init__.pyi
drubinstein: Updated it to be Optional[Union[]] instead. My bad.

PR #119, opened 2022-07-17
Add auto-generated Sphinx documentation.

### LINKED ISSUE #105: API documentation
Please publish API documentation based on docstrings, etc. and link from your README.md.  Also, in particular the `AudioFile` object is lacking docstrings.

Thanks!

psobot:
Closes #105. See https://spotify.github.io/pedalboard for a working documentation demo of what's in this PR.

The tl;dr; this PR introduces:
 - a statically compiled HTML website (in `docs/`) that can be served via GitHub Pages as our documentation site
 - a [Sphinx](https://www.sphinx-doc.org/en/master/) project (in `docs/sources`) that generates the above statically-compiled HTML website - mostly from our `.pyi` stub files and existing `.py` files _automatically_
 - a script (`scripts/generate_type_stubs_and_docs.py`) that:
   - runs our existing scripts to create `.pyi` files, generating type stubs from native code
   - runs `mypy.stubtest` on the resulting stubs
   - patches both `pybind11_stubgen` (to generate better stubs for enums) and `sphinx` (to allow it to read `.pyi` files)
   - generates the entire Sphinx static site
   - includes a `--check` mode that checks that the result matches what's in the Git repo (to be used by GitHub Actions)
 - some more detailed documentation and examples
 - fixes some tiny inconsistencies in the Pedalboard API:
   - when we return an untyped Numpy array, we now use `py::array` instead of `py::handle`
   - method overloads are (mostly) removed in favour of accepting a `py::array` and doing our own datatype `switch` statement (making both the documentation and the type hints cleaner)

TODOs for the future:
 - [ ] clean up `scripts/postprocess_type_hints.py` so that we don't need to do as much string manipulation
 - [ ] make an upstream PR to `pybind11_stubgen` to avoid the need to patch it
 - [ ] get this PR merged to allow us to use `interrogate` to measure documentation coverage: https://github.com/econchick/interrogate/pull/122


docs/_sources/compatibility.rst.txt
```diff
@@ -0,0 +1,21 @@
+Plugin Compatibility
```
drubinstein, line 1: Maybe add a piece here that if their plugin doesn't work to make an issue and maybe make an issue tag called "plugin compatibility"

drubinstein (approved):

docs/_sources/compatibility.rst.txt
psobot: Great idea!

Issue #120, opened 2022-07-21
Building on RedHat - call of overloaded convertFixedToFloat is ambiguous


rm-star:
### Expected behaviour

Building pedalboard on RedHat succeeds 

### Actual behaviour

C++ compilation fails with
`pedalboard/pedalboard/io/ReadableAudioFile.h:287:25: error: call of overloaded ‘convertFixedToFloat(float*&, const int*, float&, long long int&)’ is ambiguous
                   numSamples);
                             ^
    In file included from JUCE/modules/juce_audio_basics/juce_audio_basics.h:86,
                     from /home/wilhadden/work/misc/pedalboard/pedalboard/JuceHeader.h:25,
                     from /home/wilhadden/work/misc/pedalboard/pedalboard/python_bindings.cpp:22:
    JUCE/modules/juce_audio_basics/buffers/juce_FloatVectorOperations.h:225:31: note: candidate: ‘static void juce::FloatVectorOperations::convertFixedToFloat(float*, const int*, float, int)’
         static void JUCE_CALLTYPE convertFixedToFloat (float* dest, const int* src, float multiplier, int num) noexcept;
                                   ^~~~~~~~~~~~~~~~~~~
    JUCE/modules/juce_audio_basics/buffers/juce_FloatVectorOperations.h:227:31: note: candidate: ‘static void juce::FloatVectorOperations::convertFixedToFloat(float*, const int*, float, size_t)’
         static void JUCE_CALLTYPE convertFixedToFloat (float* dest, const int* src, float multiplier, size_t num) noexcept;
                                   ^~~~~~~~~~~~~~~~~~~
`

### Steps to reproduce the behaviour

On RedHat 8, install all dependencies
Activate virtual environment
run python3 -m pip install -e

psobot:
Closed by #121.

PR #121, opened 2022-07-21
Allow Pedalboard to compile against JUCE 6.15+ on Red Hat.

rm-star:
Problem

This is a solution for https://github.com/spotify/pedalboard/issues/120
When building on RedHat, and possibly other linux platforms, the
following error is seen during the C++ compilation

error: call of overloaded ‘convertFixedToFloat
    (float*&, const int*, float&, long long int&)’ is ambiguous

Solution

Conditionally, based on the preprocessor flag LINUX, cast the argument
numSamples to size_t

Result

Compilation of the C++ code will succeed on RedHat

### DIFF
--- a/pedalboard/io/ReadableAudioFile.h
+++ b/pedalboard/io/ReadableAudioFile.h
@@ -281,7 +281,12 @@ public:
         for (int c = 0; c < numChannels; c++) {
           juce::FloatVectorOperations::convertFixedToFloat(
               channelPointers[c], (const int *)channelPointers[c], scaleFactor,
+#ifdef LINUXX
+              static_cast<size_t>(numSamples));
+#else
               numSamples);
+#endif
+	  //https://github.com/Tencent/rapidjson/issues/873
         }
       }
     }
--- a/pedalboard/io/ReadableAudioFile.h
+++ b/pedalboard/io/ReadableAudioFile.h
@@ -281,12 +281,7 @@ public:
         for (int c = 0; c < numChannels; c++) {
           juce::FloatVectorOperations::convertFixedToFloat(
               channelPointers[c], (const int *)channelPointers[c], scaleFactor,
-#ifdef LINUXX
-              static_cast<size_t>(numSamples));
-#else
-              numSamples);
-#endif
-	  //https://github.com/Tencent/rapidjson/issues/873
+              static_cast<int>(numSamples));
         }
       }
     }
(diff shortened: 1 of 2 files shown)


pedalboard/io/ReadableAudioFile.h
```diff
@@ -281,7 +281,12 @@ class ReadableAudioFile
         for (int c = 0; c < numChannels; c++) {
           juce::FloatVectorOperations::convertFixedToFloat(
               channelPointers[c], (const int *)channelPointers[c], scaleFactor,
+#ifdef LINUXX
```
psobot, line 284: I don't see a `LINUXX` define anywhere in Pedalboard or JUCE; is this a Red Hat-only thing?

pedalboard/io/ReadableAudioFile.h
```diff
@@ -281,7 +281,12 @@ class ReadableAudioFile
...
+              static_cast<size_t>(numSamples));
```
psobot, line 285: In the version of JUCE we're currently using in Pedalboard (6.1.4), `convertFixedToFloat` seems to take an `int`. Can we change the type of `numSamples` to an `int` instead, rather than upcasting here?

pedalboard/io/ReadableAudioFile.h
```diff
@@ -281,7 +281,12 @@ class ReadableAudioFile
...
+#else
               numSamples);
+#endif
+	  //https://github.com/Tencent/rapidjson/issues/873
```
psobot, line 289:
```suggestion
```

psobot (changes requested):

pedalboard/io/ReadableAudioFile.h
rm-star: I pushed the wrong merge, apologies I will change it it LINUX which comes from setup.py

pedalboard/io/ReadableAudioFile.h
rm-star: I was trying to use the latest version of JUCE 7.0.1 which also has a size_t signature. Probably a little premature doing this.
I could change this line to 
`static_cast<int>(numSamples));`
which would mean I don't need the the preprocessor check for LINUX and this will also work for JUCE 7.x
Is that preferable? I am not keen on changing the type coming in to this function as that could have unforeseen side effects, but I could try it if you want.

psobot (approved):

PR #136, opened 2022-08-02
CONTRIBUTING: Switching pip install to use local install

rm-star:
Problem
    
    The current developer install is using
    pip setup.py
    which creates a distributable installation. If a developer makes changes
    to code then they essentailly need to build and reinstall the package
    
    Solution
    
    If pip install -e . is used then the local site packages will
    create a symbolic link which means developers can make cahnges locally
    and they will automatically picked up.
    It also has the bonus of installing dependencies automatically

    A good case for this install format is made here https://stackoverflow.com/questions/42609943/what-is-the-use-case-for-pip-install-e

    I am able to test this on linux but not Mac or Windows, though python should
    operate in the same way where ever it runs


rm-star:
I should also say that the -e in the install command means that it is an editable install. I think that helps explain it a bit better

Issue #158, opened 2022-11-02
How is tempo set? [question]


rhelsing:
I cant find anywhere in the documentation where tempo can be established. I have loaded a vst3 delay plugin and having an input tempo is critical to have the plugin perform appropriately.

psobot:
You're correct - tempo isn't provided to plugins, as Pedalboard doesn't have a concept of musical tempo (or most other musical concepts that DAWs use, for that matter).

If it's not possible to set a delay length in milliseconds with your plugin, Pedalboard could be modified to allow providing tempo when `process()` is called; [the underlying C++ framework (JUCE) seems to support providing tempo to plugins](https://github.com/juce-framework/JUCE/blob/2b16c1b94c90d0db3072f6dc9da481a9484d0435/modules/juce_audio_processors/format_types/juce_VSTPluginFormat.h#L76-L77), and plumbing that value through to Python should be possible.

psobot:
Sure! Without having dug too deeply yet, it looks like you’d have to:

- modify the `VST3Plugin` class in Pedalboard to add a `tempo_bpm` property, and default it to 120bpm
- Add an `ExtraFunctions` callback (linked above) that can provide this `tempo_bpm` value to the VST plugin when it asks for it
- wire up this property to the Python bindings so that users can set the BPM on each plugin that needs it.

There’s a different approach that could also work, but would be a bit harder: we could add a `tempo_bpm` property to the `process` method on all plugins, then have the `VSTPlugin` class receive that data when `process` is called. That would be more intuitive for users (as all plugins should probably share the same tempo), but may be a more difficult change.

PR #172, opened 2022-12-01
Add tempo support

rhelsing:
In reference to #158 

@psobot Could you let me know if I'm going in the right direction with this? My c++ knowledge is rusty.

### DIFF
--- a/pedalboard/ExternalPlugin.h
+++ b/pedalboard/ExternalPlugin.h
@@ -789,6 +789,8 @@ public:
         channelPointers[i] = outputBlock.getChannelPointer(i);
       }
 
+      // TODO: need to bind to JUCE here / set tempo?
+
       // Depending on the bus layout, we may have to pass extra buffers to the
       // plugin that we don't use. Use vector here to ensure the memory is
       // freed via RAII.
@@ -1043,6 +1045,7 @@ inline void init_external_plugins(py::module &m) {
            "Show the UI of this plugin as a native window. This method will "
            "block until the window is closed or a KeyboardInterrupt is "
            "received.");
+      // TODO: need to bind to JUCE here?
 #endif
 
 #if JUCE_PLUGINHOST_AU && JUCE_MAC
@@ -1107,6 +1110,7 @@ inline void init_external_plugins(py::module &m) {
            "Show the UI of this plugin as a native window. This method will "
            "block until the window is closed or a KeyboardInterrupt is "
            "received.");
+      // TODO: need to bind to JUCE here?
 #endif
 }
 
--- a/pedalboard/ExternalPlugin.h
+++ b/pedalboard/ExternalPlugin.h
@@ -1000,7 +1000,7 @@ inline void init_external_plugins(py::module &m) {
              return std::make_unique<ExternalPlugin<juce::VST3PluginFormat>>(
                  pathToPluginFile, pluginName, tempoBpm);
            }),
-           py::arg("path_to_plugin_file"), py::arg("plugin_name") = py::none())
+           py::arg("path_to_plugin_file"), py::arg("plugin_name") = py::none(), py::arg("tempo_bpm") = 120)
       .def("__repr__",
            [](ExternalPlugin<juce::VST3PluginFormat> &plugin) {
              std::ostringstream ss;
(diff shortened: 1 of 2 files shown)


pedalboard/pedalboard.py
```diff
@@ -677,6 +681,7 @@ def __init__(
                 )
             _VST3Plugin.__init__(self, path_to_plugin_file, plugin_name)
             self.__set_initial_parameter_values__(parameter_values)
+            self.__set_extra_functions__(tempo_bpm)
```
psobot, line 684: No need to use a separate function for this - we can add [a constructor parameter to `VST3Plugin`](https://github.com/spotify/pedalboard/blob/4bfe24e44ad1f3e5a8672308567904d06bf09c2c/pedalboard/ExternalPlugin.h#L998-L1002) such that we just pass the value through into `_VST3Plugin.__init__` a couple lines above.

rhelsing:
@psobot When you have time, I'd love to hear your feedback on this

pedalboard/pedalboard.py
```diff
@@ -565,6 +565,9 @@ def __set_initial_parameter_values__(
                 )
             setattr(self, key, value)
 
+    def __set_extra_functions__(self, tempo_bpm: int = 120):
+        tempo_bpm = self.tempo_bpm
+
```
psobot, line 570: This won't do anything: `tempo_bpm` is just a local variable. I think you probably want to pass `tempo_bpm` through as a parameter into `_VST3Plugin.__init__(self, path_to_plugin_file, plugin_name)` [on line 682](https://github.com/spotify/pedalboard/pull/172/files#diff-1b515ecfb538f3b7e79927772f8c4d03ae3b10b68e3349ab05c6eb289293305aR682) instead.

pedalboard/ExternalPlugin.h
```diff
@@ -994,9 +995,10 @@ inline void init_external_plugins(py::module &m) {
...
       "support Linux, this will fail).")
       .def(py::init([](std::string &pathToPluginFile,
-                       std::optional<std::string> pluginName) {
+                       std::optional<std::string> pluginName,
+                       std::optional<int> tempoBpm) {
              return std::make_unique<ExternalPlugin<juce::VST3PluginFormat>>(
-                 pathToPluginFile, pluginName);
+                 pathToPluginFile, pluginName, tempoBpm);
            }),
            py::arg("path_to_plugin_file"), py::arg("plugin_name") = py::none())
```
psobot, line 1003: You'll need to add `tempo_bpm` as a parameter here too, so that Python knows about it:
```suggestion
           py::arg("path_to_plugin_file"), py::arg("plugin_name") = py::none(), py::arg("tempo_bpm") = 120)
```

pedalboard/ExternalPlugin.h
```diff
@@ -994,9 +995,10 @@ inline void init_external_plugins(py::module &m) {
       "(i.e.: if you're running Linux but trying to open a VST that does not "
       "support Linux, this will fail).")
       .def(py::init([](std::string &pathToPluginFile,
-                       std::optional<std::string> pluginName) {
+                       std::optional<std::string> pluginName,
+                       std::optional<int> tempoBpm) {
```
psobot, line 999: I think we'll need to decide if `tempoBpm` can be optional or not; in your Python code below, you pass a default value of 120 BPM. Making `tempoBpm` an `std::optional` here means that it could be `None` as well; it sounds like it should be mandatory argument with a default value, rather than an optional type.

psobot (commented):
Looks like this is moving in the right direction! I don't know that this will compile, though - `getBpm` and `setBpm` aren't defined on `ExternalPlugin`, and the `ExternalPlugin` constructor doesn't take the new `tempoBpm` argument yet.

exit-rain:
Please add this functionality to the master branch of Pedalboard! We can't use delays, reverbs as well as some synths which have synchronization with host tempo.

RoyalCities:
> In reference to #158
> 
> @psobot Could you let me know if I'm going in the right direction with this? My c++ knowledge is rusty.

Just wanted to throw in my support for this. If we could somehow set a global bpm variable it would push this into overdrive and open up ALOT of options given so many vst makers default to vst host sync and don't offer any manual control

PR #175, opened 2022-12-02
Fix the build

drubinstein:
With the recent commit to remove Python3.6 support, the build has broken. This PR should fix those issues.


psobot (approved):
LGTM, thanks @drubinstein! Didn't expect `ubuntu-latest` to change out from under us and cause build-breaking issues.

PR #180, opened 2022-12-21
Remove a single tick so that a documentation link gets created correctly

juhakivekas:
Noticed the link from `load_plugin` to `VST3Plugin` was broken on the public docs


psobot (approved):
Thanks @juhakivekas!

PR #229, opened 2023-07-11
Fix bug in the vst example in README.md

DavidDiazGuerra:
Hello, this is just a fix for a small bug in the vst example of the readme. The code crashes if you don't include the name of the `sample_rate` argument.


psobot (approved):
Thanks @DavidDiazGuerra! I swapped the order of the arguments to match the order of the arguments in the function signature.

PR #234, opened 2023-07-27
Add support for instrument plugins that require event loops to run on start.

### LINKED ISSUE #233: Some VST3 instruments produce no audio output until a UI is shown.
Hello,

I've been working with some VST3i instruments in Windows and some of them do not work or have very weird behaviors. 

With Spitfire's [BBC Symphony Orchestra Discover](https://www.spitfireaudio.com/bbc-symphony-orchestra-discover), I only get any audio if I do a `plt` plot before calling the instrument. This can be replicated with the following script:

```
from pedalboard import Pedalboard, Reverb, load_plugin
from mido import Message
import matplotlib.pyplot as plt

instrument = load_plugin("C:/Program Files/Common Files/VST3/BBC Symphony Orchestra (64 Bit).vst3/Contents/x86_64-win/BBC Symphony Orchestra (64 Bit).vst3")

# plt.plot(0)
# plt.show()

audio = instrument(
  [Message("note_on", note=60), Message("note_off", note=60, time=5)],
  duration=5, # seconds
  sample_rate=44100,
)

plt.plot(audio.T)
plt.show()
```

I've run this code in one computer with Windows 11 and in one with Windows 10 and in both cases `audio` is all zeros. However, uncommenting the `plt` lines, `audio` is correctly filled with the synthesized signal. Any idea about what can be going on?

I've also tried with Orchestools' [Strings plugin](https://musictop69.wixsite.com/orchestools/orchestools-two) and neither in Windows or in Linux could I get any synthesized signal, even doing a `plt` plot before calling the instrument. 

I've tested the installation of both plugins using Reaper DAW and they work well there. The Magical8BitPlug2 plugin works well with `pedalboard` in those installations.

psobot:
Fixes #233.

The VST3 API allows plugins to assume that an event loop is always running; and as such, many plugins add callbacks to that event loop to handle asynchronous tasks like resource loading. Pedalboard doesn't pump this event loop (unless a GUI is being shown) and so these plugins may never get their callbacks, which usually results in them rendering silent audio when `process(...)` is called.

This PR adds a workaround: when loading an instrument plugin, we now perform a quick test at the end of the plugin's load function. We send a single MIDI note (channel 1, middle C, full velocity) and pump the event loop until we get audio back, or until we hit a 5-second timeout.

When tested with a variety of instrument plugins, plugin load time seems to increase by between 0ms and 250ms on my machine - but crucially, we no longer render silent audio for these plugins.


pedalboard/ExternalPlugin.h
```diff
@@ -758,6 +763,103 @@ class ExternalPlugin : public AbstractExternalPlugin {
...
 
+  /**
+   * Send a MIDI note into this plugin in an attempt to wait for the plugin to
+   * "warm up". Many plugins do asynchronous background tasks on launch (such as
+   * loading assets from disk, etc). These background tasks may depend on the
+   * event loop, which Pedalboard does not pump by default.
+   *
+   * Returns true iff the plugin rendered audio; false if no audio was received.
+   */
+  bool attemptToWarmUp(int timeoutMilliseconds = 5000) {
```
drubinstein, line 774: For the future, time values like this should be a [std::chrono::duration](https://en.cppreference.com/w/cpp/chrono/duration) object.

drubinstein (commented):
Makes sense. Can you make the warm up optional or make `timeoutMilliseconds` configurable from the Python API? 

I'm curious if you could decrease warm up time by choosing a smaller bufferSize or sampleRate.

PR #236, opened 2023-08-01
Return an integer for AudioFile.samplerate when possible.

psobot:
Pedalboard internally represents sample rates as floating-point numbers; however, the vast majority of sample rates used in production are integers (like 44100, 48000, etc).

This PR changes the `AudioFile` class to return an integer for the `samplerate` property if the fractional part of the sample rate is 0. This will make it possible to do things like passing `f.samplerate` to the `f.read` method directly, without needing to cast to `int` first.

This PR also makes it possible to pass a floating-point number to `AudioFile.read`. An exception will be thrown if the provided number has a fractional part.


pedalboard_native/io/__init__.pyi
```diff
@@ -319,7 +319,9 @@ class ReadableAudioFile(AudioFile):
         """
         Close this file, rendering this object unusable.
         """
-    def read(self, num_frames: int = 0) -> numpy.ndarray[typing.Any, numpy.dtype[numpy.float32]]:
+    def read(
+        self, num_frames: typing.Union[float, int] = 0
```
drubinstein, line 323: I'm a little worried about this change. What does it mean for pedalboard to return a fractional sample?

psobot: See the comment below in the docstring:

```
.. note::
    For convenience, the ``num_frames`` argument may be a floating-point number. However, if the
    provided number of frames contains a fractional part (i.e.: ``1.01`` instead of ``1.00``) then
    an exception will be thrown, as a fractional number of samples cannot be returned.
```

PR #249, opened 2023-09-08
Add pedalboard.time_stretch for time stretching.

psobot:
Pedalboard already includes [Chris Cannam's excellent Rubber Band](https://github.com/breakfastquay/rubberband) library for [its `PitchShift` plugin](https://spotify.github.io/pedalboard/reference/pedalboard.html#pedalboard.PitchShift). This PR adds a tiny utility function, called `pedalboard.time_stretch`, to expose Rubber Band's other useful bit of functionality: time stretching.

Note that as time stretching changes the length of the audio, this functionality is exposed as a separate standalone function, rather than as a `Plugin` subclass, and as such **cannot be included in `Pedalboard` effects chains**. However, getting time stretching "for free" by adding ~128 lines of glue code and tests seems pretty worthwhile.


pedalboard/TimeStretch.h
```diff
@@ -0,0 +1,88 @@
...
+/*
+ * A wrapper around Rubber Band that allows calling it independently of a plugin
+ * context, to allow for both pitch shifting and time stretching on fixed-size
+ * chunks of audio.
+ *
+ * The `Plugin` base class requires that one sample of audio output is always
+ * provided for every sample input, but this assumption does not hold true
+ */
+static juce::AudioBuffer<float>
+timeStretch(const juce::AudioBuffer<float> input, double sampleRate,
```
rabitt, line 33: nit: this function is called timeStretch, but looks like it can do time stretching + pitch shifting at once?

pedalboard/TimeStretch.h
```diff
@@ -0,0 +1,88 @@
...
+            double speedFactor, double pitchShiftInSemitones) {
```
rabitt, line 34: nit: call this `stretchFactor`? Speed changes are clasically resampling, i.e. pitch and time at the same time, which is distinct from "stretch" which is only time.

rabitt (approved):
a couple of naming nits, otherwise looks good!

f0k:
@psobot: Thanks a lot for doing this, very useful! In extending some augmentation pipeline, I noticed that the `time_stretch()` function is slower and gives different results than the `PitchShift` plugin:
```
In [1]: import numpy as np

In [2]: import pedalboard

In [3]: x = np.sin(2 * np.pi * 0.001 * np.arange(250*512)).astype(np.float32)

In [4]: %time y1 = pedalboard.time_stretch(x, 44100, pitch_shift_in_semitones=2)
CPU times: user 371 ms, sys: 15.3 ms, total: 386 ms
Wall time: 385 ms

In [5]: %time y2 = pedalboard.Pedalboard([pedalboard.PitchShift(semitones=2)])(x, 44100)
CPU times: user 193 ms, sys: 0 ns, total: 193 ms
Wall time: 193 ms

In [6]: import matplotlib.pyplot as plt

In [7]: plt.plot(y1.squeeze()[:1000])
Out[7]: [<matplotlib.lines.Line2D at 0x7ff160723460>]

In [8]: plt.plot(y2.squeeze()[:1000])
Out[8]: [<matplotlib.lines.Line2D at 0x7ff160738160>]
```
plot

Off the top of your head, do you know the reason? Setting `high_quality=False` gives a similar speed to `PitchShift`, but yet another result (a third one). I'd just like to understand if I'm missing out on anything when switching from `PitchShift` to `time_strech()` (for the use case when I actually need both stretching and shifting).

PR #285, opened 2023-12-19
Fix for the bug in JUCE submodule

gloriajjl:
JUCE submodule has a bug in the code caused by unneeded optimization in VST3PluginInstance::releaseResources method. This causes tests failures on the plug-ins on our end. This fix overrides juce_VST3PluginFormat.cpp file.


psobot (approved):
Looks great, thank you!

Issue #277, opened 2023-11-18
Cannot load .fxp file for Serum


turian:
I am trying to load a factory FXP preset for Serum:

```
synth_plugin = load_plugin("/Library/Audio/Plug-Ins/Components/Serum.component")
synth_plugin.load_preset("/Library/Audio/Presets/Xfer Records/Serum Presets/Presets/Plucked/PL Crusty Pluck [LCV].fxp")
```

The problem is that `synth_plugin` is `AudioUnitPlugin`, not `VST3Plugin`, and `AudioUnitPlugin` doesn't have load_preset method.

If instead I try:
```
synth_plugin = load_plugin("/Library/Audio/Plug-Ins/VST/Serum.vst")
```

I get this error:
```
  File "/opt/homebrew/anaconda3/envs/faceswap/lib/python3.10/site-packages/pedalboard/_pedalboard.py", line 781, in load_plugin
    raise ImportError(
ImportError: Failed to load plugin as VST3Plugin or AudioUnitPlugin. Errors were:
	VST3Plugin: Unable to load plugin /Library/Audio/Plug-Ins/VST/Serum.vst: unsupported plugin format or load failure.
	AudioUnitPlugin: Unable to load plugin /Library/Audio/Plug-Ins/VST/Serum.vst: unsupported plugin format or load failure. macOS requires plugin files to be moved to /Library/Audio/Plug-Ins/Components/ or ~/Library/Audio/Plug-Ins/Components/ before loading.
```

How can I load a particular Serum preset in pedalboard, to get a particular instrument sound?

0xdevalias:
The following issues may be relevant here:

- https://github.com/spotify/pedalboard/issues/245
- https://github.com/spotify/pedalboard/issues/187

zfarrell13:
@turian did you ever find a way to do this?  ive been trying for the past few days with all of the methods and workarounds from above but no luck with serum sepcifically

0xdevalias:
> did you ever find a way to do this?

These 2 PR's look super useful, assuming one or the other (or both) eventually lands:

- #289 
- #297

Ma5onic:
it's possible that serum and its presets are in vst2 format.

Example of loading vst2 presets:
```c++
bool PluginProcessor::loadPreset(const std::string& path) {
  THROW_ERROR_IF_NO_PLUGIN

  try {
    if (!std::filesystem::exists(path)) {
      throw std::runtime_error("File not found: " + path);
    }

[...]

Issue #257, opened 2023-10-05
VST plugin Program presets not accessible


petrikaj:
I have tried to access and change iZotope's VST plugins' Program presets but it seems they are not exposed.

[Parameters](https://github.com/spotify/pedalboard/issues/217) work ok.

Here's what MrsWatson prints out from plugin —display-info
 
<img width="322" alt="mrswatson" src="https://github.com/spotify/pedalboard/assets/52623440/7da39a2b-42e2-4e58-a181-daf02e6b3b95">

Notice it's not part of the parameters.

Using plugin.show_editor() allows to access those presets and changing them will change the parameters accordingly after quitting the showeditor(/).

MrsWatson allows to change it while loading the plugin:
"mrswatson64 --plugin 'plugin_name,preset_name’ -i input.wav -o output.wav"

0xdevalias:
Looking at the MrsWatson source code, this seems to be where the 'Programs' is being printed:

- https://github.com/teragonaudio/MrsWatson/blob/32ec2b4404190d7f2710cdd9beaf4b153e6ff37d/source/plugin/PluginVst2x.cpp#L686-L699

It reads from:

- `data->pluginHandle->numPrograms`
- `data->pluginHandle effGetProgramNameIndexed` `nameBuffer->data`
- `data->pluginHandle effGetProgramName` `nameBuffer->data`

Googling for `effGetProgramNameIndexed`, it seems to be part of the VST2 SDK or similar.

Eg. Here's a go package that implements bindings for it:

- https://github.com/pipelined/vst2
  - > Build vst2 plugins and hosts with Go
  - > vst2 implements VST2 SDK API. It provides capabilities to build both VST2 hosts and plugins.
  - https://pkg.go.dev/github.com/pipelined/vst2
    - ```
      // EffGetProgramNameIndexed passed to get program name by index.
	  // Index: program index.
	  // Ptr: *[maxProgNameLen]uint8 buffer for program name.
	  // Return: true for success.
	  EffGetProgramNameIndexed
      ```

Pedalboard doesn't seem to support VST2, but not sure if there is an equivalent function in VST3. We can start by looking at the VST3 SDK:

- https://github.com/steinbergmedia/vst3sdk
  - > VST 3 Plug-In SDK

Searching for `effGetProgramNameIndexed` doesn't seem to turn up any results there (neither does `GetProgramName` / `ProgramName`)

Not sure if that's because the concept no longer exists in VST3, or if it just has a different name/etc.

[...]

Issue #245, opened 2023-08-19
Loading non .vstpreset files with load_preset function


gonzaloarca:
I'm currently trying to load the [Mini V3 plugin](https://www.arturia.com/products/software-instruments/mini-v/overview) from the Arturia V Collection using the Pedalboard library and change presets programatically. I can load the plugin correctly, but for some reason I'm not able to load the factory presets. 

The presets are not in .vstpreset format, and I actually believe they use some proprietary format from Arturia.

## Library versions
- Python: 3.11.4 (main, Jun 15 2023, 07:29:58) [Clang 14.0.3 (clang-1403.0.22.14.1)]
- Pip: 23.0.1
- Pedalboard: 0.7.8

## Steps to reproduce
```python
from pedalboard import load_plugin 
mini = load_plugin('/Library/Arturia/Mini V3/Mini V3.vst3')

mini.load_preset("/Library/Arturia/Presets/Mini V3/Factory/Factory/24 VCO Unisson")
```

## Output
```python
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
Cell In[68], line 1
----> 1 mini.load_preset("/Library/Arturia/Presets/Mini V3/Factory/Factory/24 VCO Unisson")

RuntimeError: Plugin returned an error when loading data from preset file: /Library/Arturia/Presets/Mini V3/Factory/Factory/24 VCO Unisson
```

## Preset file structure
Here's what the 24 VCO Unisson preset file looks like in plain text:
```
[...]

Issue #187, opened 2023-01-08
Save and Load Presets Automatically for VST3 Plugins?


Isak-Andersson:
Hello. I'm trying to save the changes I make after I use the show_editor function. What's the best way of saving presets for plugins?

I tried dumping the hole session with dill.dump_session but this gives me this warning and error:
`PicklingWarning: Pickling a PyCapsule (None) does not pickle any C data structures and could cause segmentation faults or other memory errors when unpickling.
TypeError: cannot pickle 'pedalboard_native.io.WriteableAudioFile' object`

Using load_preset with .fxp files isn't much better. It doesn't work with certain plugins and I also don't see a way of saving it. Maybe a save_preset function could be a feature in future versions?
  
Trying to load a preset for ChorusGAS.vst3 gives me this error:
`RuntimeError: Plugin returned an error when loading data from preset file`

Any ideas on how to approach this issue?

psobot:
Hi @Isak-Andersson!

There are a couple ways forward here, in order from easiest to hardest:
 - Any changes you make with `show_editor` _should_ be exposed to Pedalboard via the parameters of the plugin itself. (i.e.: if your plugin has a `frequency_hz` parameter, changing that value in the plugin's UI should result in the corresponding value changing when accessing `plugin.frequency_hz`).

    A simple way to serialize this data would be something like:
    ```python
    my_plugin = load_plugin(...)
    my_plugin.show_editor() # make some edits

    # Read out the value of every parameter that this plugin exposes:
    param_value_dict = {parameter_name: getattr(my_plugin, parameter_name) for parameter_name in my_plugin.parameters.keys()}

    # Do something with param_value_dict, like serializing it to JSON:
    with open("params.json", "w") as f:
        json.dump(param_value_dict, f)

    # To reload, just iterate over this dictionary and use `setattr` instead:
    for parameter_name, serialized_value in param_value_dict.items():
        setattr(my_plugin, parameter_name, serialized_value)
    ```
 - Alternatively, if your plugin doesn't expose all of its UI parameters to its host, this would be a good reason for us to add setters and getters for _all_ plugin state; i.e.: `my_plugin.state` could return a `bytes` object containing the entire plugin's serialized state, ready to be reloaded later.

Let me know if that first option works for you at all, and if not, I'll see what I can do about adding a `.state` property.

Isak-Andersson:
Thanks for the answer. I'm not very familiar with JSON files. Is there a good reason to use them instead of just pure txt files?

[...]

PR #289, opened 2024-02-11
Addition: Implementing Save and Load Functionality for VST3 Plugin States

Myuqou:
### Problem:

- Some VST3 plugins don't expose all UI parameters to the host, hindering 
programmatic manipulation without manual intervention. This is evident in 
plugins like Native Instrument's Kontakt and Plogue's Sforzando, where 
instrument or effect selection is confined to the editor window, posing 
challenges for automated loading.

- Additionally, issues arise with plugins featuring presets incompatible with 
existing load methods, or being overly complicated in setting their parameters, 
as evidenced in previous issue reports:
#187
#245
#257
#277).

### Solution:

- Implemented save_plugin_state and load_plugin_state methods in ExternalPlugin 
to address the above issues. Leveraging JUCE's getStateInformation and 
setStateInformation, these methods facilitate saving and loading plugin 
configurations to/from disk in much the same way that a DAW would.

- The save_plugin_state method also makes a copy of the accessible parameters 
which the load_plugin_state reapplies twice, similar to how Pedalboard's 
reinstantiatePlugin method works.

- The binary state information is converted to a base64 string using JUCE, and 
the recorded parameters are also converted to a string. The two data types are 
combined, being separated by a pipe character, which is not used in the base64 
encoding. The resulting character string is then saved to disk.

Using the new methods in Python is as simple as this example:

```
plugin = load_plugin('path/to/plugin/plugin.vst3')
plugin.show_editor()   // Manually configure the plugin to the desired state
plugin.save_plugin_state("path/to/your/directory/plugin.sav")
```

Later, to restore the plugin to the previously saved state:

`plugin.load_plugin_state("path/to/your/directory/plugin.sav")`

- Note that the extension on the save file does not have to be .sav, the user 
can name the file whatever they want.  The data is just a character string 
stored in what is essentially a text file.

### Results:

I tested the new methods on the following VST3 plugins and they work flawlessly:
- Plogue's Sforzando
- Native Instrument's Kontact 7
- Native Instrument's Guitar Rig 7
- Blue Cat Free Amp
- Vital (the wavetable synth)

I do not have access to Serum, so I can't verify that this will address 
Issue #277, but Vital is the OSS counterpart of Serum and I was able to load a 
previously saved state, which included instrument presets, without any issue.

### Potential Problems:

- I am a beginner level, self taught programmer who has no experience working 
on team projects, so I don't know if I followed all of the best practices when 
writing this code.  Reviewers should keep that in mind in case I messed 
something up.

- I noticed that Pedalboard's reinstantiatePlugin method goes through a number 
of steps when loading a state back into the plugin after resetting it. I only 
incorporated the part where, after the state is set, the parameters are 
reassigned twice.  I didn't run into any problems, but I am not familiar with 
the situations that reinstantiatePlugin is trying to address so someone with 
more experience need to verify that there isn't going to be an issue there.

- The names of the parameter keys do not update if you switch instruments on an 
existing plugin instance by loading a plugin state.  I am not sure how to fix 
this, but a best practice would be to load the parameter settings on a fresh 
instance of the plugin.

- There is currently no safeguard or error warning when if a user tries to load 
a state into the wrong plugin.  It's currently up to the user to keep their 
saved state files organized.  I am not sure how to fix this at this time.

- Unfortunately, I couldn't test this on a Mac, so I can't guarantee 
compatibility with AU plugins. In theory, it should work, but I don't want to 
implement a feature I can't test.  Updating the code for AU plugins should be 
as simple as adding two lines for the pybind11 bindings.

- I did not add any type hints or updates to the documentation as I don't feel 
entirely confident about getting it right without causing more confusion.

### DIFF
--- a/pedalboard/ExternalPlugin.h
+++ b/pedalboard/ExternalPlugin.h
@@ -31,6 +31,8 @@
 
 #include "process.h"
 
+#include <fstream>
+
 #if JUCE_MAC
 #include <AudioToolbox/AudioUnitUtilities.h>
 #endif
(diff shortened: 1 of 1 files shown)


Myuqou:
As tempted as I am to take on the challenge, it would probably be a better idea for you to make the changes if you have the time as you are more familiar with the program.

I wasn't sure if capturing the parameters was even a necessary step as in every test I conducted I was able to simply save and load the state information without running into any problems.  If the copy of the parameters is not needed then that would make it much easier to just grab the state information and pass it back to Python.

0xdevalias:
> @psobot Let me know if you'd like to make these changes or if you'd prefer me to do so; I'd be happy to make these changes in this PR if you like.

> @Myuqou As tempted as I am to take on the challenge, it would probably be a better idea for you to make the changes if you have the time as you are more familiar with the program.

@psobot Curious if you were still up for making the changes required to land this? Would be an awesome feature to have in `pedalboard`!

0xdevalias:
I was reading a bit more about saving/loading state for VST3, and it sounds like there are 2 different kinds of state:

- https://www.kvraudio.com/forum/viewtopic.php?t=597225
  - > Vst3 has 2 sets of states, a processor state, which is the state of the audio processor (all parameter values etc) , and a controller state, which is the state of the gui and everything that is unique to that and not part of the audio processing.
    > 
    > Iirc, a host calls `Processor::SetState()` with the processor state and then `EditController::SetComponentState()` with the same state, and finally `EditController::SetState()` with the controller state. Maybe not exactly in that order, but you get the idea.
- https://steinbergmedia.github.io/vst3_dev_portal/pages/Technical+Documentation/API+Documentation/Index.html
  - > A VST 3 audio effect or instrument basically consists of two parts: a processing part and an edit controller part. The corresponding interfaces are:
    > 
    > - Processor: [Steinberg::Vst::IAudioProcessor](https://steinbergmedia.github.io/vst3_doc/vstinterfaces/classSteinberg_1_1Vst_1_1IAudioProcessor.html) + [Steinberg::Vst::IComponent](https://steinbergmedia.github.io/vst3_doc/vstinterfaces/classSteinberg_1_1Vst_1_1IComponent.html)
    > - Controller: [Steinberg::Vst::IEditController](https://steinbergmedia.github.io/vst3_doc/vstinterfaces/classSteinberg_1_1Vst_1_1IEditController.html)
  - https://steinbergmedia.github.io/vst3_dev_portal/pages/Technical+Documentation/API+Documentation/Index.html#persistence
    - > ## Persistence
      > The host stores and restores the complete state of the processor and of the controller in project files and in preset files:
      > 
      > - [Steinberg::Vst::IComponent::getState](https://steinbergmedia.github.io/vst3_doc/vstinterfaces/classSteinberg_1_1Vst_1_1IComponent.html#a10db03106be8ba89d23859fa6be5d9f6) + [Steinberg::Vst::IComponent::setState](https://steinbergmedia.github.io/vst3_doc/vstinterfaces/classSteinberg_1_1Vst_1_1IComponent.html#a77ac39bcc5c4b15818b1a87de2573805) store and restore the DSP model.
      > - [Steinberg::Vst::IEditController::getState](https://steinbergmedia.github.io/vst3_doc/vstinterfaces/classSteinberg_1_1Vst_1_1IEditController.html#a10db03106be8ba89d23859fa6be5d9f6) + [Steinberg::Vst::IEditController::setState](https://steinbergmedia.github.io/vst3_doc/vstinterfaces/classSteinberg_1_1Vst_1_1IEditController.html#a77ac39bcc5c4b15818b1a87de2573805) store and restore any GUI settings that are not related to the processor (like scroll positions etc).
      > - Restore: When the states are restored, the host passes the processor state to both the processor and the controller ([Steinberg::Vst::IEditController::setComponentState](https://steinbergmedia.github.io/vst3_doc/vstinterfaces/classSteinberg_1_1Vst_1_1IEditController.html#a4c2e1cafd88143fda2767a9c7ba5d48f)). A host must always pass that state to the processor first. The controller then has to synchronize its parameters to this state (but must not perform any [IComponentHandler](https://steinbergmedia.github.io/vst3_doc/vstinterfaces/classSteinberg_1_1Vst_1_1IComponentHandler.html) callbacks). After restoring a state, the host rescans the parameters (asking the controller) in order to update its internal representation.
      
Am I right in assuming that previously we were able to access the 'processor state' (param values), and as of #297 we can now access the 'controller state'?

PR #291, opened 2024-02-26
Release the GIL when reading from BytesIO objects.

psobot:
Prior to this PR, Pedalboard used the standard Python [`IOBase`](https://docs.python.org/3/library/io.html#io.IOBase) interface to read audio data from file-like objects, by calling their `seek`, `seekable`, `read`, and `tell` methods. However, these method calls happen in Python, and require the `ReadableAudioFile` class to take the GIL for every read operation.

This PR adds an optimization just for the [`BytesIO`](https://docs.python.org/3/library/io.html#io.BytesIO) class: if passed a `BytesIO` object (or any object with a `.getbuffer()` method that returns a [`memoryview`](https://docs.python.org/3/c-api/memoryview.html)), Pedalboard will now automatically call `.getbuffer()` and store a pointer to the underlying bytes, rather than a Python object reference to the `BytesIO` object itself.

This allows reading audio from `BytesIO` objects without holding the GIL, which dramatically increases performance in multi-threaded applications.


pedalboard/io/PythonInputStream.h
```diff
@@ -32,20 +32,35 @@ bool isReadableFileLike(py::object fileLike) {
...
+
+    return {};
+  }
+}
+
 /**
  * A juce::InputStream subclass that fetches its
  * data from a provided Python file-like object.
  */
 class PythonInputStream : public juce::InputStream, public PythonFileLike {
```
drubinstein, line 59: Not sure if you intend to inherit from this class again, but if you don't then [you can add the final specifier](https://en.cppreference.com/w/cpp/language/final).

pedalboard/io/PythonInputStream.h
```diff
@@ -219,4 +234,63 @@ class PythonInputStream : public juce::InputStream, public PythonFileLike {
...
+
+  virtual std::string getRepresentation() override { return repr; }
+
+  virtual bool isSeekable() noexcept override { return true; }
+
+  virtual juce::int64 getTotalLength() noexcept override { return totalLength; }
+
+  virtual int read(void *buffer, int bytesToRead) noexcept override {
+    int start = offset;
+    int end = std::min(offset + bytesToRead, totalLength);
```
drubinstein, line 271: Nit: It may be easier to read if this is instead written as 
```c++ 
if (offset + bytesToRead >= totalLength) {
  bytesToRead = totalLength - offset;
}
```
Then the subtraction logic below doesn't need repeating and you don't need start. I sometimes get a little confused with `end` because even though `end` is usually exclusive, occasionally it isn't.

pedalboard/io/PythonInputStream.h
```diff
@@ -219,4 +234,63 @@ class PythonInputStream : public juce::InputStream, public PythonFileLike {
...
+    }
+  }
+
+  virtual std::string getRepresentation() override { return repr; }
+
+  virtual bool isSeekable() noexcept override { return true; }
+
+  virtual juce::int64 getTotalLength() noexcept override { return totalLength; }
+
+  virtual int read(void *buffer, int bytesToRead) noexcept override {
```
drubinstein, line 269: You're mixing camel case and snake case here. In much of the rest of the code you use snake case for variable naming, shouldn't the same apply in this class?

pedalboard/io/AudioFileInit.h
```diff
@@ -155,15 +155,24 @@ inline void init_audio_file(
           "__new__",
           [](const py::object *, py::object filelike, std::string mode) {
```
drubinstein, line 156: Doesn't apply to memoryviews and bytesIO, but [pybind11 will convert pathlib.Path objects to std::filesystem](https://pybind11.readthedocs.io/en/stable/advanced/cast/overview.html?highlight=path#list-of-all-builtin-conversions).

Though I am a little confused why you can't specialize this function on `py::buffer`

drubinstein (approved):
Mostly c++ style changes.

pedalboard/io/AudioFileInit.h
psobot: Specializing on `py::buffer` would work, but that would add another overload to [the already-fairly-dense documentation for the constructor](https://spotify.github.io/pedalboard/reference/pedalboard.io.html#pedalboard.io.AudioFile): image

We already rewrite this type for the `.pyi` files and documentation as `typing.BinaryIO` - this question is a good reminder that I should change that type to `typing.BinaryIO | memoryview` to be more exact about what we accept in the docs.

pedalboard/io/PythonInputStream.h
drubinstein: [Below](https://github.com/spotify/pedalboard/pull/291/files#diff-b41e5e23d58e8432233614cd153f7a1ebeab13d11d0c953141fdc1e7ea3a9266R292) I was looking at the definition for `py::buffer_info` and overfit to it. Fair point.

PR #297, opened 2024-03-01
Add binary state field to VST3 and AU plugins

kmontag:
This adds a read-only `state` field and a `set_state` method to external plugins, similar to what's been discussed in https://github.com/spotify/pedalboard/issues/187.

This is intended as a more robust alternative to saving and loading parameter values. The value is retrieved/updated using JUCE's `getStateInformation`/`setStateInformation`.

Since it stores the plugin's entire native state (within a JUCE wrapper structure), the blob can also be used to generate preset files in other formats, e.g. `vstpreset`.


kmontag:
Somehow I missed https://github.com/spotify/pedalboard/pull/289 before submitting this :) 

I'll leave this PR open as it's a bit of a different approach, but feel free to close it if the other approach is preferred.

0xdevalias:
From #289:

> I have a couple high-level suggestions before I think we can merge this:
>
> - These new methods look like they work just fine - but they require the user to save the plugin state to a file. Pedalboard's design philosophy is to integrate with Python as much as possible; in keeping with that philosophy, it might be more intuitive to change these save/load methods into a [`.def_property`](https://pybind11.readthedocs.io/en/stable/classes.html#instance-and-static-fields), so the plugin's state becomes a property that users can do what they like with:
> 
>     ```python
>     # Instead of:
>     my_plugin.save_plugin_state("to/filename")
> 
>     # ...we could use a property:
>     len(my_plugin.state) # 12,345 bytes
>     with open("my_filename.sav", "wb") as f:
>         f.write(my_plugin.state)
> 
>     # ...which also allows people to do what they want with the
>     # plugin state, without requiring them to write to disk:
>     my_plugin.state = some_function(my_plugin.state)  # modify the plugin's internal state
>     ```
> 
> _Originally posted by @psobot in https://github.com/spotify/pedalboard/issues/289#issuecomment-1959721455_

---

From a quick skim of this PR, it looks like it's closer (`def_property_readonly` + `def` `set_state` ) to the suggested method (`def_property`).

I'm not sure if the 'separated read/write' approach in this PR is better than just a pure `def_property`.

kmontag:
Thanks all! The changes look great, glad to see this land.

@psobot 🤘 🤘

PR #305, opened 2024-03-14
Add channel mapping support

dodeca-6-tope:
Map channels to their respective speakers for supported audio formats. This change solves an issue where receivers can't determine how to map channels for multi channel audio files.

### DIFF
--- a/pedalboard/io/WriteableAudioFile.h
+++ b/pedalboard/io/WriteableAudioFile.h
@@ -356,9 +356,19 @@ public:
     }
 
     juce::StringPairArray emptyMetadata;
-    writer.reset(format->createWriterFor(outputStream.get(), writeSampleRate,
-                                         numChannels, bitDepth, emptyMetadata,
-                                         qualityOptionIndex));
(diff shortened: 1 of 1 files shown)


dodeca-6-tope:
@psobot a more flexible approach would be to take an optional layout and fallback to the changes I introduced here if not provided.

dodeca-6-tope:
@psobot not that I know of, but you can verify this change by writing and playing a multichannel track on your receiver if you have one lying around. Audacity also fills up the channels with the correct signals unlike before. The first thing I did is to look for a visualizer but I haven't found any.

dodeca-6-tope:
Tested on a receiver that initially mapped a down-mixed version (to mono) of the file to all channels equally and after this change it mapped as expected.

dodeca-6-tope:
Then it must be an issue with the receiver itself I guess, will try again tomorrow on a different device.

dodeca-6-tope:
Might be an issue with the receiver, will try again tomorrow with another device.

PR #316, opened 2024-04-22
Speed up tests to ~3 minutes.

psobot:
Also introduces some changes to reduce test flakiness on CI.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -10,3 +10,6 @@
 [submodule "vendors/libgsm"]
 	path = vendors/libgsm
 	url = https://github.com/timothytylee/libgsm.git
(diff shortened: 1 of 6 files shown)


.gitmodules
```diff
@@ -10,3 +10,6 @@
 [submodule "vendors/libgsm"]
 	path = vendors/libgsm
 	url = https://github.com/timothytylee/libgsm.git
+[submodule "vendors/pybind11"]
```
drubinstein, line 13: Apparently you can [pip install pybind11](https://pybind11.readthedocs.io/en/stable/installing.html#include-with-pypi)

PR #347, opened 2024-06-25
Allow processing square buffers.

### LINKED ISSUE #336: Any `Plugin` cannot process stereo buffers containing only 2 samples
## Problem

I'm developing a DSP graph using `pedalboard`. It basically consist of reading arbitrary audio files at configurable chunk size, processing them with a `Chain` and then writing back the output to file system. While writing unit tests I hit this puzzling behaviour that prevents me from moving forward.

If any `Plugin.process(...)` implementation receives a buffer that has shape `(2, 2)`, it will fail with

`RuntimeError: Unable to determine channel layout from shape!`

This is not really an uncommon scenario, and can even be replicated with the [first example](https://spotify.github.io/pedalboard/examples.html#quick-start) provided in the Pedalboard website, by having a stereo `some-file.wav` at sample rate `x` that is exactly `x+2` samples long. In general, if any chosen buffer reading size happens to cause a reminder of 2 samples over the audio file being read, the square matrix condition is met and no processing can be done in pedalboard°.

## Stack trace

Inspecting the `pedalboard` source code, I recreated the C++ call stack, which is not shown in Python (most recent call last):

- `py::array_t<float> process(...)` defined in `process.h`

https://github.com/spotify/pedalboard/blob/273eb538193b6e3557c92745a6dbbf6f1d627481/pedalboard/process.h#L260-L261

- `py::array_t<float> processFloat32(...)` defined in `process.h`

https://github.com/spotify/pedalboard/blob/273eb538193b6e3557c92745a6dbbf6f1d627481/pedalboard/process.h#L166

- `template <typename T> ChannelLayout detectChannelLayout(...)` defined in `BufferUtils.h`

https://github.com/spotify/pedalboard/blob/273eb538193b6e3557c92745a6dbbf6f1d627481/pedalboard/BufferUtils.h#L44-L45

Specifically, the last line is hit whenever `Plugin.process(...)` is called with a buffer that has shape `(n, n)` (i.e. a square matrix).

## Proposed solution

**Is it possible to change the function signature of `Plugin.process(...)` so that it can accept an explicit `ChannelLayout`**? The semantics would be:
- if `None` (default when not provided), then autodetect (existing behaviour, backwards compatible)
- if provided, then skip autodetection (thus, accepts 2x2 buffers)

Since `ReadableAudioFile.read()` reliably returns non-interleaved buffers, then to me it makes sense to enforce the same layout when processing, leaving auto-detection for other scenarios / backwards compatibility?

I can help implementing this if needed 👍

° _the workaround would be to check the buffer and if squared then append an extra padding sample to make it rectangular, but this seems less than ideal and unnecessarily complicates client code._

psobot:
Fixes #336. (cc @mttbernardini)

This PR fixes the issue that occurs when passing a square (i.e.: `(x, x)`) buffer into a `Plugin` instance. Like https://github.com/spotify/pedalboard/pull/338, this PR adds caching of the last-detected channel layout to each `Plugin` instance, making the following example pass:

```python3
board = Pedalboard([Reverb(), Gain()])
with AudioFile("file_with_17_stereo_samples.wav") as f:
    while f.tell() < f.frames:
        # Read three frames at a time; but the last frame will
        # end up with only two samples in it.
        output = board(f.read(3), f.samplerate)
```

Prior to this PR, the above example would have failed as the last `f.read(3)` call would return an array of shape `(2, 2)`.


mttbernardini:
Hey @psobot, thank you for fixing this. Does this PR include the "memorization" feature as in the `AudioFile` scenario, i.e. process `(n, 0)` or `(0, n)` to hint the layout for the next buffer call?

PR #351, opened 2024-07-05
Add VST3 `preset_data` property to get/set .vstpreset data directly

kmontag:
Setting the property has the same behavior as `load_preset` (but doesn't require a file). Getting the property returns the .vstpreset data representing the current plugin state.

Previously, it wasn't quite possible to generate .vstpresets purely from `pedalboard` objects, since they need to include a [class ID](https://forum.juce.com/t/how-to-get-vst3-class-id-aka-cid-aka-component-id/41041/3), which isn't exposed anywhere else.

_(I'm running into issues getting type stubs generated - perhaps someone could generate stubs before merging?)_

### DIFF
--- a/tests/test_external_plugins.py
+++ b/tests/test_external_plugins.py
@@ -320,14 +320,10 @@ def test_get_vst3_preset(plugin_filename: str):
     ), "Preset data for {plugin_filename} is not in .vstpreset format"
     # Check that the class ID (8 bytes into the data) is a 32-character hex string.
     cid = preset_data[8:][:32]
-    assert all(
-        c in b"0123456789ABCDEF" for c in cid
-    ), f"CID contains invalid characters: {cid}"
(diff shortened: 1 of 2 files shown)


tests/test_external_plugins.py
```diff
@@ -302,6 +306,59 @@ def test_preset_parameters(plugin_filename: str, plugin_preset: str):
         ), f"Expected attribute {name} to be different from default ({default}), but was {actual}"
 
 
+@pytest.mark.skipif(
+    not AVAILABLE_VST3_PLUGINS_IN_TEST_ENVIRONMENT,
+    reason="No VST3 plugin containers installed in test environment!",
```
psobot, line 311:
```suggestion
    reason="No VST3 plugins installed in test environment!",
```

kmontag:
Awesome thanks!

Issue #243, opened 2023-08-17
Raw Value Setter Prbolem


kleblackford:
Setting the raw_value attribute of an AudioProcessorParameter object does not work. I was attempting this (suggested [here](https://github.com/spotify/pedalboard/blob/a5500d2b66dc2967fe5699f56f108c44884a0889/pedalboard/_pedalboard.py#L235C1-L235C1)) as a workaround for the fact that resolution is lost when setting a parameter. 

### Expected behaviour

Setting the raw_value attribute changes the plugin parameter value directly.

### Actual behaviour

Nothing happens.

### Steps to reproduce the behaviour

The following code with the attached VST plugin demonstrates this problem.

[Amplifier.zip](https://github.com/spotify/pedalboard/files/12370195/Amplifier.zip)

```
import pedalboard

amplifier = pedalboard.load_plugin("Amplifier.vst3")

print("Parameters:", amplifier.parameters)

amplifier.amplifier_gain.raw_value = 0.5

print("Parameters:", amplifier.parameters)
```

Output:

```
Parameters: {'amplifier_gain': <pedalboard.AudioProcessorParameter name="Amplifier Gain" discrete raw_value=0.01 value=1.00 range=(0.0, 100.0, 0.1)>, 'bypass': <pedalboard.AudioProcessorParameter name="Bypass" discrete raw_value=0 value=Off boolean ("False" and "True")>}
Parameters: {'amplifier_gain': <pedalboard.AudioProcessorParameter name="Amplifier Gain" discrete raw_value=0.01 value=1.00 range=(0.0, 100.0, 0.1)>, 'bypass': <pedalboard.AudioProcessorParameter name="Bypass" discrete raw_value=0 value=Off boolean ("False" and "True")>}
```

nafeu:
I was able to reproduce your issue and I have also been attempting to use `raw_value` to set a parameter value, however when looking through the docs I found that it isn't exactly specified that `raw_value` could be used as a setter, it only makes mention of accessing that value and implies "Raw Control". I think this means you can use still use a value between [0, 1] but you have to calculate the actual value to set based on the given bounds. As a sort of workaround I did the following:

I used the `range` on `effect.parameters[PARAM_NAME]` to find the min-max values, they usually look something like:

```
(0.0, 100.0, 0.1)
```

Where the first two values are the min and max respectively. Then I added these two helpers:

```
def map_value(value, out_min, out_max):
  return out_min + value * (out_max - out_min)

[...]

PR #367, opened 2024-08-02
Fix `raw_value` setter for parameters of external plugins 

fblang:
Setting the `raw_value` property of `AudioProcessorParameter` objects previously didn't have an effect, see #243.
I implemented the fix proposed by @kleblackford, adding a `__setattr__` method to `WeakTypeWrapper`.


fblang:
@psobot Is this in a good state or does it still need work? Happy to make any adjustments if needed!

Issue #366, opened 2024-08-01
AudioStream: proposal to enable support for Linux


sarmentow:
I'd like to propose enabling Linux support for the AudioStream class in Pedalboard. As a user, I was excited to discover Pedalboard's capabilities for real-time audio processing, but was disappointed to find that Linux support had been removed.

After digging into the history, I found that Linux support was removed due to issues with the ALSA dependency (see #98 and commit b5561ab). However, I was able to get the AudioStream example working on Ubuntu 22.04 by removing the ifndef guards and making some changes to the audioDeviceIOCallback method.

Here's an example of code that currently does not work in Linux and that works after my modifications:
```python3
from pedalboard import Pedalboard, Compressor, Reverb
from pedalboard.io import AudioStream

# Open up an audio stream:
stream = AudioStream(
  input_device_name="Playback/recording through the PulseAudio sound server",
  num_input_channels=2,
  num_output_channels=2,
  allow_feedback=True,
  buffer_size=128,
  sample_rate=44100,
  output_device_name="Playback/recording through the PulseAudio sound server"
) 

stream.plugins = Pedalboard([
  Reverb(),
  Compressor()
])

print(stream)
stream.run()
```
I'd like to propose merging these changes, as they seem to work for me. Are there any outstanding issues with the ALSA dependency that I'm not aware of? If not, I'd be happy to open a pull request to re-enable Linux support for AudioStream.

Let me know if this proposal is feasible and if there's anything else I can do to help!

psobot:
Wow, thanks @sarmentow!

> Are there any outstanding issues with the ALSA dependency that I'm not aware of?



> If not, I'd be happy to open a pull request to re-enable Linux support for AudioStream.

sarmentow:
Just opened the pull request for it, would be glad to receive your feedback! I took the liberty of also adding an example of using AudioStream for live audio monitoring.

psobot:
Closed by #368.

PR #368, opened 2024-08-04
Add pedalboard.io.AudioStream support for Linux

sarmentow:
Implements #366


sarmentow:
Sure. I'll add the JUCE_ALSA flag and set it to 0 in the compiler flags for MacOS and Windows to avoid including anything ALSA-related (should've done that before, sorry!), but I'm curious as to what should be done in the Linux build. 

I'm thinking of adding back the ifdefs you had for preventing compilation of the AudioStream header (and the juce_audio_devices header) and only include these files and link ALSA if some environment variable like LINUX_ENABLE_ALSA is set, similar to how debug builds work (i.e. by setting the DEBUG environment variable).

Tell me if that sounds good and I'll start work on it right away!

sarmentow:
Okay, I'll investigate the possibility of statically linking `libasound` and come back with the results.

sarmentow:
@psobot I managed to statically link `libasound` in linux builds by adding the `alsa-lib-dev` package in the before-all install command in the `pyproject.toml`. 

Locally building `cp310-manylinux_x86_64`, the size of the wheel is 4.5mb for the 0.9.12 version, versus 4.1mb for the same version listed in PyPI (so ~400kb increase in size).

Do you think that's reasonable?

sarmentow:
Yes, the example using `AudioStream` works as expected after the changes when installing Pedalboard from the new wheel. I'll commit these changes and wait for CI results.

sarmentow:
@psobot Pre-build on linux is still failing with the error: `'alsa/asoundlib.h' file not found`
:confused:. I didn't realize that I also need to add the `alsa-lib` package (in Ubuntu it's under the name `libasound2-dev`) to the `Install Linux Dependencies` step in the workflow file. Will commit this change as well.

sarmentow:
It appears that now that AudioStream actually loads on Linux, the tests in `test_audio_stream.py` must be reconsidered. We can simply delete the `test_create_stream_fails_on_linux` and remove all `@pytest.mark.skipif(platform.system() == "Linux", ...)` and it should work.

sarmentow:
I've removed the Linux-specific skips from the AudioStream tests, but encountered errors during local test runs when attempting to create AudioStream objects with certain output devices :thinking:. To rule out any potential issues with my local environment, I've verified that the same errors occur when running the tests locally on the master branch. Given this, I'm inclined to believe that the issue is specific to my machine, and I'd like to re-run the tests in CI to confirm.

sarmentow:
While reading the logs of failed checks and reading on the workflows themselves, we seem to be exceeding the 6 hour time limit, which is causing tests to fail on Linux. I'll look into what may be causing tests taking this long.

Also, the step in which libasound is removed is getting a weird error regarding unmet dependencies regarding the Java Runtime? Will try looking into that as well.

sarmentow:
@psobot Great news! I was able to reproduce the bug that's been causing our tests to hang 😊.

The issue arises when there are no audio devices available in Linux environments. In these cases, `AudioStream.default_output_device_name` and `AudioStream.default_input_device_name` return empty strings instead of `None`. Creating an `AudioStream` with both devices set to `None` leads to a runtime error - which avoids hanging - however, the code doesn't currently handle empty strings, resulting in an assertion failure in JUCE (`juce_AudioDeviceManager.cpp:804`) and subsequent hanging.

I'll add this consideration to validate constructor arguments in `AudioStream`

sarmentow:
After some reading, I found the `snd-dummy` driver (see: https://www.alsa-project.org/wiki/Matrix:Module-dummy) which provides a virtual sound card for testing AudioStream on Linux. Now a couple of test devices should show up when testing in CI.

sarmentow:
@psobot pinging for CI approval :heart_hands: :

sarmentow:
@psobot Thank you for all the help and support! This was my first PR in a bigger project, definitely learned a lot.  👍👍

PR #381, opened 2024-11-04
Don't segfault on zero-channel MP3s.

### LINKED ISSUE #379: Pedalboard 0.9.16 segfaults on MP3 (attached)
Hi,

Pedalboard 0.9.16 segfaults without error on the attached MP3 (*) - it processed some 70k other mp3 files perfectly fine. System is MacOS Sonoma. Here is a minimal example:

```
from pedalboard.io import AudioFile

with AudioFile('099134.mp3') as f:
    audio = f.read(f.frames)
```

(*) The MP3 came from the Free Music Archive but there was some kind of decompression error which produced this file - this context should not matter but who knows.
[099134.zip](https://github.com/user-attachments/files/17586966/099134.zip)

psobot:
Fixes #379.


tdhopper (approved):
ship it

Issue #287, opened 2024-01-16
Actual resampled audio length != expected resampled audio length


daniel-deychakiwsky:
I am not getting back the expected audio length in the target sample rate but I do w/ resampy (sinc_window). When I provide a 44_100 source sampled signal, I do get back the expected. When I provide a 48_000 source sampled signal, pedalboard is one sample shy. Any idea why this may be? It is an issue in the code because function expects to get back the expected # of samples under the target sample rate. Here's a snip to reproduce it:

```python
import io
import numpy as np
import soundfile as sf
from pedalboard import Resample
from pedalboard.io import AudioFile, StreamResampler

# equals target_samples
s_sr = 44_100
# does not equal target_samples!
# s_sr = 48_000
t_sr = 22_050
dur = 5

source_samples = s_sr * dur
target_samples = int(np.floor((source_samples / s_sr) * t_sr))

noise = np.random.rand(source_samples).astype(np.float32)

audio_filelike = io.BytesIO()

with sf.SoundFile(
    audio_filelike,
    "w",
    channels=1,
    samplerate=s_sr,
    format="WAV",
    subtype="PCM_16",
) as f:
    f.write(noise)
    
audio_filelike.seek(0)

audio = np.copy(sf.SoundFile(audio_filelike, "r").read(dtype="float32"))

resampler = StreamResampler(
    source_sample_rate=s_sr,
    target_sample_rate=t_sr,
    num_channels=1,
    quality=Resample.Quality.WindowedSinc
)
resampled = np.concatenate([resampler.process(audio), resampler.process(None)], axis=1).squeeze()

print(resampled.shape, target_samples)
```

hyperc54:
Hi!
I confirm I can reproduce, I would be happy to have a look at fixing!

PR #385, opened 2024-11-10
tests: Fix `input_signal` variable in `test_returned_sample_count` being too short

hyperc54:
## Problem

I believe the way the `input_signal` variable (in the `test_returned_sample_count` test) is currently defined on master may contain a typo.
I would expect `input_signal` to represent a realistic input that would contain 1s worth of samples for different sample rates.
Currently, it always contain 3 samples, which makes the different chunk sizes testing mostly worthless.

## Solution

Switching the order of parameters when calling `np.linspace` creates a more realistic input (mostly, a longer array).
I kept using np.linspace instead of switching to np.rand (that might be more realistic?) because `np.linspace` is probably faster to run.

## Result

The test is still green, but it now more accurately covers/tests all different chunk_sizes scenarios.

## Additional context

I was working on fixing issue #287 (which I believe I may have a fix for) and I wanted to amend the `returned_sample_count` test to make it cover the problem highlighted in #287. The way np.linspace was called surprised me so I first want to make sure Im understanding this properly :)

### DIFF
--- a/tests/test_stream_resampler.py
+++ b/tests/test_stream_resampler.py
@@ -181,7 +181,7 @@ def test_flush(sample_rate: float, target_sample_rate: float, quality: Resample.
 def test_returned_sample_count(
     sample_rate: float, target_sample_rate: float, chunk_size: int, quality
 ) -> np.ndarray:
-    input_signal = np.linspace(0, sample_rate, 3, dtype=np.float32)
+    input_signal = np.linspace(0, 3, num=int(sample_rate), dtype=np.float32)
     resampler = StreamResampler(sample_rate, target_sample_rate, 1, quality)
 
     print("input", input_signal)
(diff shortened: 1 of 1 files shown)


tests/test_stream_resampler.py
```diff
@@ -169,7 +169,7 @@ def test_flush(sample_rate: float, target_sample_rate: float, quality: Resample.
 def test_returned_sample_count(
     sample_rate: float, target_sample_rate: float, chunk_size: int, quality
 ) -> np.ndarray:
-    input_signal = np.linspace(0, sample_rate, 3, dtype=np.float32)
+    input_signal = np.linspace(0, 3, num=int(sample_rate), dtype=np.float32)
     resampler = StreamResampler(sample_rate, target_sample_rate, 1, quality)
 
     print("input", input_signal)
```
hyperc54, line 175: This may need to go too?

PR #398, opened 2025-02-10
Add missing py.typed for pedalboard_native

matangover:
Add missing `py.typed` in `pedalboard_native` to solve mypy import errors.

### Details

Before this fix, mypy doesn't recognize any imports that come from pedalboard_native (e.g. `pedalboard.io.AudioFile`).

To reproduce, run:
```shell
mypy -c "from pedalboard.io import AudioFile"
```
You get the error: `Module "pedalboard.io" has no attribute "AudioFile"`. If you add `--follow-untyped-imports`, the error goes away:
```shell
mypy -c "from pedalboard.io import AudioFile" --follow-untyped-imports
```

After this fix, `--follow-untyped-imports` is no longer necessary.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -6,5 +6,6 @@ recursive-include tests *.py
 graft pedalboard
 graft JUCE
 graft tests
+include pedalboard_native/py.typed
 
 exclude build-info.yaml service-info.yaml mkdocs.yml


psobot (approved):
This is excellent, thanks so much Matan!

I think we might also need to add this file to `MANIFEST.in` so that it gets packaged when we deploy - I'll push a commit to fix that.

Issue #402, opened 2025-02-26
Feature Request: Fuzz Plugin for Pedalboard


LorenzoMonti:
Hello Pedalboard team,

I really appreciate the work you’ve done on this library—it's a fantastic tool for audio processing! I’d love to see a fuzz plugin added to the available effects.

Why?

- Fuzz is a staple effect in many music genres, especially rock and experimental electronic music.
-  Distortion and overdrive are common, but fuzz has a distinct, aggressive character that can't be easily replicated with other effects.
- This would enhance Pedalboard’s flexibility for guitarists and producers looking for vintage or experimental tones (this is also my case :D).

Potential Implementation Ideas:
 
- A simple fuzz model with a gain and tone control.
- Possibly an emulation of classic fuzz circuits (e.g., RAT, Fuzz Face, Big Muff).

Would this be something the team is open to adding? I'm already working on it.
Thanks for considering this feature!

anguzo:
Surprisingly Distortion with 40db kind of sounds like fuzz (I am not a professional musician so maybe my judgement is not on spot). The work of Distortion and Clipping is confusing to me as from description I assumed- that they are kind of soft and hard clipping, but at high db Distortion also notably increases perceived loudness while Clipping does not? This should not be characteristic of any of them but begs the need for normalization after effect application.

PR #403, opened 2025-02-28
Feature request: add Fuzz plugin for Pedalboard (#402)

LorenzoMonti:
Problem

Fuzz is a staple effect in many music genres, particularly rock and 
experimental electronic music. While distortion and gain effects exist 
in Pedalboard, fuzz has a unique aggressive character that cannot be 
easily replicated. Adding a fuzz plugin would expand Pedalboard’s 
flexibility, providing more options for guitarists and producers 
seeking vintage or experimental tones. 

This feature request was initially discussed in #402.

Solution

Implemented a new Fuzz effect, emulating a classic fuzz pedal. 
The effect consists of:
- A two-stage clipping process:
  - Hard diode clipping with a threshold of 0.25.
  - Soft clipping using a tanh function.
- A tone control stage, implemented as a low-pass filter.
- Unit tests verifying expected behavior.

Result

Pedalboard now includes a fuzz effect, broadening its range of 
distortion-based effects. Users can apply fuzz to their audio chains 
for a distinctive saturated sound characteristic of classic fuzz pedals.

### DIFF
--- a/pedalboard/plugins/Fuzz.h
+++ b/pedalboard/plugins/Fuzz.h
@@ -1,0 +1,112 @@
+/*
+ * pedalboard
+ * Copyright 2021 Spotify AB
+ *
+ * Licensed under the GNU Public License, Version 3.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    https://www.gnu.org/licenses/gpl-3.0.html
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+ #include <pybind11/pybind11.h>
+ #include <pybind11/stl.h>
+ #include <cmath>
+ 
+ namespace py = pybind11;
+ 
+ #include "../JucePlugin.h"
+ #include <juce_dsp/juce_dsp.h>
+ 
+ namespace Pedalboard {
+ 
+ template <typename SampleType>
+ class Fuzz
+     : public JucePlugin<juce::dsp::ProcessorChain<
+           juce::dsp::Gain<SampleType>,              // Drive stage
+           juce::dsp::WaveShaper<SampleType>,          // Hard diode clipping stage
+           juce::dsp::WaveShaper<SampleType>,          // Soft clipping stage (tanh)
+           juce::dsp::IIR::Filter<SampleType>          // Tone control (low-pass filter)
+           >> {
+ public:
+   Fuzz() : driveDecibels(static_cast<SampleType>(25)),
+            toneHz(static_cast<SampleType>(800)) {}
+ 
+   void setDriveDecibels(const float f) noexcept { driveDecibels = f; }
+   float getDriveDecibels() const noexcept { return driveDecibels; }
+ 
+   void setToneHz(const float f) noexcept { toneHz = f; }
+   float getToneHz() const noexcept { return toneHz; }
+ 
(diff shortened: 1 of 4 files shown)


pedalboard/plugins/Fuzz.h
```diff
@@ -0,0 +1,112 @@
...
+       };
+ 
+     // Third stage: Tone control via low-pass filter
+     auto coeffs = juce::dsp::IIR::Coefficients<SampleType>::makeLowPass(spec.sampleRate, toneHz);
+     this->getDSP().template get<filterIndex>().coefficients = coeffs;
+   }
+ 
+ private:
+   SampleType driveDecibels;
+   SampleType toneHz;
```
psobot, line 81: We probably don't want to keep a copy of these variables here, as they can differ from the variables set in the `ProcessorChain` depending on when `prepare` is called. (And indeed; if you use `Fuzz` in an `AudioStream`, it should be possible to call `.drive_db = x` and the audio should update without needing to stop and start the stream again.)

pedalboard/plugins/Fuzz.h
```diff
@@ -0,0 +1,112 @@
...
+           juce::dsp::WaveShaper<SampleType>,          // Hard diode clipping stage
+           juce::dsp::WaveShaper<SampleType>,          // Soft clipping stage (tanh)
+           juce::dsp::IIR::Filter<SampleType>          // Tone control (low-pass filter)
+           >> {
+ public:
+   Fuzz() : driveDecibels(static_cast<SampleType>(25)),
+            toneHz(static_cast<SampleType>(800)) {}
+ 
+   void setDriveDecibels(const float f) noexcept { driveDecibels = f; }
+   float getDriveDecibels() const noexcept { return driveDecibels; }
```
psobot, line 42:
```suggestion
   void setDriveDecibels(const float f) noexcept { this->getDSP().template get<gainIndex>().setGainDecibels(f); }
   float getDriveDecibels() const noexcept { return this->getDSP().template get<gainIndex>().getGainDecibels(f); }
```

pedalboard/plugins/Fuzz.h
```diff
@@ -0,0 +1,112 @@
...
+       };
+ 
+     // Second stage: soft clipping via tanh
+     this->getDSP().template get<shaperIndex>().functionToUse =
+       [](SampleType x) -> SampleType {
+         return static_cast<SampleType>(std::tanh(x));
+       };
+ 
+     // Third stage: Tone control via low-pass filter
+     auto coeffs = juce::dsp::IIR::Coefficients<SampleType>::makeLowPass(spec.sampleRate, toneHz);
```
psobot, line 75: Could try using `ArrayCoefficients` here to allow updating the `toneHz` parameter without requiring a realloc/extra `prepare` call here - with the caveat that I haven't worked with that API in a long time and am not sure if it would actually work here.

pedalboard/plugins/Fuzz.h
```diff
@@ -0,0 +1,112 @@
+/*
+ * pedalboard
+ * Copyright 2021 Spotify AB
```
psobot, line 3:
```suggestion
 * Copyright 2025 Spotify AB
```

psobot (commented):
Excellent work, thank you @LorenzoMonti! I've left a couple comments inline (mostly about touching the actual DSP parameters directly in the setters/getters rather than copies) - but otherwise, this is virtually a perfect pull request. Great job!

LorenzoMonti:
> Excellent work, thank you @LorenzoMonti! I've left a couple comments inline (mostly about touching the actual DSP parameters directly in the setters/getters rather than copies) - but otherwise, this is virtually a perfect pull request. Great job!

Thanks for the thorough and speedy comment! I’m still getting familiar with JUCE, so I really appreciate the guidance. I’ll go through your comments carefully and deepen my understanding of how the `ProcessorChain` works.

PR #408, opened 2025-03-19
Raise exceptions if passing int-like data into `process`.

### LINKED ISSUE #407: Passing integer-valued samples out of range should raise a descriptive error.
In at least three cases (#406, #390, #369) users have reported passing int16-scaled data as float32 without scaling it down first, which results in some scale-invariant plugins working just fine (like Reverb, Delay, etc) and others failing completely (Compressor, Limiter, PitchShift).

We can detect this with a tiny performance penalty in the `render` call, and raise an exception if the input provided is:
1. All integer values (no fractional parts)
2. Looks like it was provided by taking 16-bit integer data and just converting to float32

This would prevent this class of issue from being raised.

tsreeram1:
I am a newcomer just trying to help out :) Let me know if I am trying to solve this issue in the wrong place or made any mistakes!

Issue #407: Passing integer-valued samples out of range should raise a descriptive error

### DIFF
--- a/pedalboard/process.h
+++ b/pedalboard/process.h
@@ -174,6 +174,44 @@ processFloat32(const py::array_t<float, py::array::c_style> inputArray,
   juce::AudioBuffer<float> ioBuffer =
       copyPyArrayIntoJuceBuffer(inputArray, {inputChannelLayout});
 
+  // ======= DETECTION FOR INT16-CONVERTED FLOAT DATA =======
+  bool allIntegers = true;
+  bool looksLikeInt16 = true;
+
+  for (int c = 0; c < ioBuffer.getNumChannels(); c++) {
+    float *channelData = ioBuffer.getWritePointer(c);
+    for (int i = 0; i < ioBuffer.getNumSamples(); i++) {
+      float value = channelData[i];
+
+      // Check if the value is very close to an integer
+      if (std::abs(value - std::round(value)) > 1e-6) {
+        allIntegers = false;
+      }
+
+      // Check if the value is within int16 range
+      if (value < -32768.0 || value > 32767.0) {
+        looksLikeInt16 = false;
+      }
+
+      // If both conditions are already violated, break early
+      if (!allIntegers && !looksLikeInt16) {
+        break;
+      }
+    }
+    if (!allIntegers && !looksLikeInt16) {
+      break;
+    }
+  }
+
+  // If all values are integers or look like int16, raise an error
+  if (allIntegers) {
+    throw std::runtime_error("Error: Input contains only integer values. Ensure proper floating-point conversion.");
+  }
+  if (looksLikeInt16) {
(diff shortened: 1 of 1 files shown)


pedalboard/process.h
```diff
@@ -174,6 +174,44 @@ processFloat32(const py::array_t<float, py::array::c_style> inputArray,
...
+  bool allIntegers = true;
+  bool looksLikeInt16 = true;
+
+  for (int c = 0; c < ioBuffer.getNumChannels(); c++) {
+    float *channelData = ioBuffer.getWritePointer(c);
+    for (int i = 0; i < ioBuffer.getNumSamples(); i++) {
+      float value = channelData[i];
+
+      // Check if the value is very close to an integer
+      if (std::abs(value - std::round(value)) > 1e-6) {
```
psobot, line 187: We might want to use [`if (f == std::trunc(f))`](https://en.cppreference.com/w/cpp/numeric/math/trunc) here instead, as the use case that we're trying to detect is when a user passes in an `int16` (or, rarely, an `int32` array containing 24-bit integer data) and PyBind11 automatically casts it to float. In such cases, the values should be _exactly_ integers, as [`float` can represent all 24-bit integers exactly](https://stackoverflow.com/a/3793950/679081).

pedalboard/process.h
```diff
@@ -174,6 +174,44 @@ processFloat32(const py::array_t<float, py::array::c_style> inputArray,
...
+      }
+    }
+    if (!allIntegers && !looksLikeInt16) {
+      break;
+    }
+  }
+
+  // If all values are integers or look like int16, raise an error
+  if (allIntegers) {
+    throw std::runtime_error("Error: Input contains only integer values. Ensure proper floating-point conversion.");
```
psobot, line 208: We don't need `Error: ` as a prefix here; the user will already see this prefixed with `RuntimeError:` when it gets raised.
```suggestion
    throw std::runtime_error("Provided input contains only integer values. Pedalboard uses floating-point audio on the range [-1, 1]. Ensure that your audio is in floating point format.");
```

pedalboard/process.h
```diff
@@ -174,6 +174,44 @@ processFloat32(const py::array_t<float, py::array::c_style> inputArray,
...
+    for (int i = 0; i < ioBuffer.getNumSamples(); i++) {
+      float value = channelData[i];
+
+      // Check if the value is very close to an integer
+      if (std::abs(value - std::round(value)) > 1e-6) {
+        allIntegers = false;
+      }
+
+      // Check if the value is within int16 range
+      if (value < -32768.0 || value > 32767.0) {
```
psobot, line 192: I think we need another check here to ensure that at least _some_ values are outside of `[-1, 1]`, or else this will trigger on all-silent inputs.

psobot (changes requested):
Thanks for this @tsreeram1! I appreciate the contribution. I've triggered the CI build and left a couple comments inline. We also need to add tests before merging this.

Let me know if you're interested in updating the PR and adding tests; if not, I'm happy to take a stab at it.

PR #429, opened 2025-06-06
Fix pybind submodule url (ssh -> https)

katyo:


hyperc54 (approved):
Hi!
thanks for the PR
I don't see a specific reason why this would use SSH over HTTPS and the other submodules use HTTPS so this LGTM

katyo:
@hyperc54 Any chance that PR will be merged soon?

PR #436, opened 2025-09-29
Fix for reading certain Ogg files on Linux via file-like objects.

psobot:
This PR fixes a subtle bug where certain Ogg Vorbis files can fail to decode when passed to Pedalboard via Python file-like objects, particularly on Linux. 

The bug stems from an interaction between three layers of code:

1. When Pedalboard reads audio from a Python file-like object (i.e.: a network stream, custom wrapper, etc), it calls Python methods like `read()`, `seek()`, and `tell()`.
2. These Python objects can internally call other native C/C++ code that sets [the thread-local `errno` variable](https://en.wikipedia.org/wiki/Errno.h) as part of normal operation, but this error is not enough to cause the Python wrapper(s) to throw exceptions.
3. The `libvorbis` decoder (specifically in [`vorbisfile.c`](https://github.com/juce-framework/JUCE/blob/ddaa09110392a4419fecbb6d3022bede89b7e841/modules/juce_audio_formats/codecs/oggvorbis/libvorbis-1.3.7/lib/vorbisfile.c#L79)) checks `errno` after I/O operations to detect errors:
   ```c
   errno=0;  // Clear errno before the read
   // ... perform read operation ...
   if(bytes==0 && errno)return(-1);  // Treat errno!=0 as an error
   ```

This results in valid Ogg Vorbis files failing to parse, as - if that last `libvorbis` read returns 0 bytes, but some other code set `errno` - then the entire `read` call fails. This can be spurred on by appending a couple null bytes to the end of the file, which I do in test here.

### DIFF
--- a/tests/test_io.py
+++ b/tests/test_io.py
@@ -1592,3 +1592,68 @@ def test_write_two_by_two_buffer_with_hint():
         f.write(np.zeros((2, 2), dtype=np.float32))
 
         assert f.tell() == 2
+
+
+class ErrnoTriggeringFileLike:
+    """A file-like object that triggers errno to be set through real failed operations."""
+
+    def __init__(self, wrapped_file):
+        self._file = wrapped_file
+        self.name = getattr(wrapped_file, "name", None)
+        # Create a path that definitely doesn't exist for triggering ENOENT
+        self._nonexistent_path = str(pathlib.Path("definitely_does_not_exist_12345.txt"))
+
+    def _trigger_errno(self):
+        # Try to open a non-existent file, which sets errno to ENOENT (2)
+        # We catch the exception but errno remains set
+        try:
+            open(self._nonexistent_path, "rb")
+        except FileNotFoundError:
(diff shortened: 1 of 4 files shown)


tests/test_io.py
```diff
@@ -1592,3 +1592,68 @@ def test_write_two_by_two_buffer_with_hint():
...
+    returning to codec code.
+    """
+
+    buf = io.BytesIO()
+    buf.name = f"test.{format_name}"
+    with pedalboard.io.AudioFile(buf, "w", 44100, 2, format=format_name) as af:
+        af.write(np.random.rand(2, 44100))
+
+    # Add some random garbage past the end of the file to trigger the bug:
+    buf.write(b"\x00\x00")
```
psobot, line 1655: This was the only reliable way I could find to repro the bug in test, apart from including a proprietary audio file as part of the tests. Trailing null bytes should not prevent the file from being parsed.

PR #447, opened 2026-01-12
Add support for reading WAV files that contain MP3 data.

psobot:
We have the ability to decode MP3 files and the ability to decode WAV files, but not the ability to parse WAV files that contain MP3 data. This is supported by [the WAV spec](https://en.wikipedia.org/wiki/WAV) but is very rare.

This PR adds a wrapper `PatchedWavAudioFormat` that:
 - Checks the headers of the WAV file to detect what compression is used therein
 - If MP3 is detected, delegate to our existing MP3 parser
 - If another unsupported format is detected, raise a descriptive error message

Includes fixture tests.


pedalboard/juce_overrides/juce_PatchedWavAudioFormat.h
```diff
@@ -0,0 +1,239 @@
...
+      return useDefaultReader();
+    }
+
+    // Skip the size field
+    sourceStream->readInt();
+
+    // Check for WAVE identifier
+    if (sourceStream->readInt() != chunkName("WAVE")) {
+      return useDefaultReader();
+    }
```
hyperc54: For my understanding: we would always expect valid WAV files to contain the RIFF header, or the WAVE identifier (right?), but in case we find such a corrupted audio, we still return the default reader = JUCE to make handle it?

hyperc54 (approved):
neat!
the test error is a bit mysterious

PR #449, opened 2026-01-13
Add support for reading WAV files containing ADPCM, A-law, µ-law, or 64-bit float data.

psobot:
Does what it says on the tin; ADPCM, A-law, and µ-law are all different compression types commonly stored within WAV files, but we don't currently support them (nor do we support 64-bit floating-point WAV files).

This PR vendors the excellent [`dr_wav`](https://github.com/mackron/dr_libs/blob/master/dr_wav.h) library from @mackron to handle reading these additional formats.

### DIFF
--- a/pedalboard/juce_overrides/juce_PatchedWavAudioFormat.h
+++ b/pedalboard/juce_overrides/juce_PatchedWavAudioFormat.h
@@ -44,13 +44,14 @@ enum class WavFormatTag : unsigned short {
 };
 
 /**
- * An AudioFormatReader that uses dr_wav to decode ADPCM audio on-the-fly.
+ * An AudioFormatReader that uses dr_wav to decode audio formats not natively
+ * supported by JUCE: ADPCM (MS and IMA), A-law, µ-law, and 64-bit float.
  * This streams the audio data rather than loading it all into memory.
  */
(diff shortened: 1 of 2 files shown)


pedalboard/juce_overrides/juce_PatchedWavAudioFormat.h
```diff
@@ -39,12 +43,178 @@ enum class WavFormatTag : unsigned short {
...
 };
 
+/**
+ * An AudioFormatReader that uses dr_wav to decode ADPCM audio on-the-fly.
+ * This streams the audio data rather than loading it all into memory.
+ */
+class DrWavAudioFormatReader : public AudioFormatReader {
+public:
+  DrWavAudioFormatReader(InputStream *stream)
+      : AudioFormatReader(stream, "ADPCM WAV"), inputStream(stream) {
```
caravanuden: the reader is passed "ADPCM WAV" as its format name, but it now handles more than just ADPCM. we could consider a more generic name like "dr_wav" or dynamically set it based on the actual format?

psobot: Fixed! I'd set this dynamically, but tbh this string is never shown to the user anyways.

PR #460, opened 2026-02-10
Allow non-blocking editor window

scottmudge:
I had a particular use case which required opening the plugin's editor window in a non-blocking mode, allowing the main thread to continue doing other tasks.

There may have been some distinct purpose for it to be blocking, but it seems to work fine, at least on Linux. Perhaps certain OS'es wouldn't like it, but it's nice to have the option available.

I also changed it so that the editor window is centered on the screen when it is shown/created, rather than at the top left corner. Sort of unrelated, but I think it's more intuitive to have it centered.

### DIFF
--- a/pedalboard/ExternalPlugin.h
+++ b/pedalboard/ExternalPlugin.h
@@ -19,6 +19,7 @@
 
 #include <mutex>
 #include <optional>
+#include <atomic>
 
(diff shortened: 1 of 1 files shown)


HRudyPlayZ:
+1 for this merge, it would make it a lot easier to work with external plugins.

Issue #321, opened 2024-05-01
VST3 effect not getting applied 


Unmesh28:
- When I use Mac the effect gets applied to input audio 
- When tried with Windows, I can see the parameters of VST3 plugin and output gets generated but effect doesn't get applied. Input and output audio are same

HN026:
#476

PR #476, opened 2026-04-09
Fix VST3 effect plugins rendering dry audio on Windows

HN026:
Fixes a Windows-specific VST3 hosting issue where effect plugins could load and expose parameters but still render dry audio. This switches normal external-plugin processing to realtime mode and adds a regression test to verify that a non-bypassed VST3 effect actually changes the signal.


HN026:
#321

PR #494, opened 2026-07-07
Add `num_threads` to `AudioFile::encode`.

psobot:
This PR adds a `num_threads` parameter to `AudioFile::encode`, which - **for MP3 only**:
 - splits the provided audio buffer into separate chunks
 - encodes each chunk in a separate parallel thread (releasing the GIL)
 - properly concatenates the resulting chunks to produce a valid and artifact-free MP3

In practice, this allows us to create MP3 files with a slightly sub-linear speedup (i.e.: using 2 threads goes 1.9x faster, 8 threads goes 6.5x faster). For CBR, the resulting file is nearly identical in size; for VBR, the lack of a usable bit reservoir inflates file sizes by up to 10%.


tdhopper (approved):
Manually tested and verified speed claims

PR #510, opened 2026-09-09
Raise TypeError for malformed MIDI messages

### LINKED ISSUE #489: normalize_midi_messages() silently drops malformed messages instead of raising an error

Hashim1999164:
Fixes #489

Malformed MIDI events were dropped with no error. The caller just got a shorter list.

normalize_midi_messages now raises TypeError with the index and the bad value.

## Tests
* One element tuple missing a timestamp
* Bare bytes

### DIFF
--- a/pedalboard/midi_utils.py
+++ b/pedalboard/midi_utils.py
@@ -45,7 +45,7 @@ def normalize_midi_messages(_input) -> List[Tuple[bytes, float]]:
     into a juce::MidiBuffer on the C++ side.
     """
     output = []
-    for message in _input:
+    for index, message in enumerate(_input):
         if hasattr(message, "bytes") and hasattr(message, "time"):
             output.append((bytes(message.bytes()), message.time))
         elif (isinstance(message, tuple) or isinstance(message, list)) and len(message) == 2:
@@ -57,6 +57,10 @@ def normalize_midi_messages(_input) -> List[Tuple[bytes, float]]:
             elif not isinstance(message, bytes):
                 message = bytes(message)
             output.append((message, time))
+        else:
+            raise TypeError(
+                f"Unable to interpret MIDI message at index {index}: {message!r}"
+            )
 
     # Detect the case in which the provided timestamps
     # are likely delta values rather than absolute values:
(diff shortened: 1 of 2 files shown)


cjacoby-sptfy:
Almost exact duplicate of #507

Hashim1999164:
oh yeah didnt see 507. all good then

PR #511, opened 2026-09-17
Fix Linux wheel illegal-instruction crashes

### LINKED ISSUE #454: Pedalboard 0.9.21 causes 'Illegal Instructions' on linux
Pedalboard 0.9.21 causes an 'Illegal Instructions' error to appear after importing the module, specifically on my linux installation. Most likely a numpy issue that was already solved in newer versions, but I'm not sure?

Image

cjacoby-sptfy:
## Summary

Fixes #454 by preventing Linux x86_64 wheels from accidentally requiring CPU instructions newer than the project’s documented baseline.

The manylinux build previously used `-march=native`. Because wheel builders may run on newer CPUs, this could compile host-specific instructions into the distributed wheel. On older CPUs, importing `pedalboard` could then terminate with `SIGILL` (`Illegal instruction`).

## Changes

  - Default Linux x86 builds to the AVX baseline using `-mavx`.
  - Preserve native optimization as an explicit opt-in with `USE_MARCH_NATIVE=1`.
  - Explicitly disable native optimization in wheel-building CI.
  - Add a compatibility smoke test for manylinux x86_64 wheels under QEMU CPU models representing:
    - Intel Ivy Bridge
    - AMD EPYC Milan
  - Verify that the installed wheel can import `pedalboard` and process audio with `Gain`.
  - Document the Linux x86_64 AVX requirement.

### Native GCP machines

The exact manylinux x86_64 wheels produced by this PR’s CI run were tested on native GCP VMs using:

- Intel Ivy Bridge
- AMD Milan

All CPython wheels from 3.10 through 3.15 successfully imported and processed audio on both machines: **12/12 passed**.

As a negative control, the published `pedalboard==0.9.21` CPython 3.11 wheel reproduced the original failure on both machines:

```text
Illegal instruction (core dumped)
exit=132
```

This confirms that the test environments can detect the original incompatibility.

### CI and local testing

- All 36 wheel-building CI jobs passed.
- The new QEMU compatibility check passed for CPython 3.10–3.15 manylinux x86_64 wheels.
- Full local Linux test suite:
    - 26,761 passed
    - 50 skipped

The only non-passing job in the CI run was the macOS Python 3.15 test job, which timed out after six hours. It did not report a test failure and does not exercise the Linux wheel-building path changed here.


.github/workflows/all.yml
```diff
@@ -160,7 +160,7 @@ jobs:
           CCACHE_IGNOREOPTIONS: "-fno-strict-overflow -fwrapv -W* -arch x86_64 arm64 -dynamic -fno-common -g -I/usr/local/opt/*"
           CCACHE_LOGFILE: ${{ github.workspace }}/.ccache_log
           CCACHE_DEBUG: "1"
-          USE_PORTABLE_SIMD: "1" # Use portable SIMD to avoid "Illegal instruction" on different runner CPUs
```
caravanuden, line 163: now that `USE_PORTABLE_SIMD` is fully removed, anyone with that var in their build scripts will silently get `-mavx` (I think this is fine?). is it worth a note in the release/changelog so folks know to update their build scripts to `USE_MARCH_NATIVE=1` if they want native optimizations?

scripts/test_linux_wheel_cpu_compatibility.py
caravanuden, line 1: would it be worth adding a negative test with a model that does not support avx, that asserts the smoke test fails?

caravanuden (approved):
looks good! just a couple questions

tdhopper (approved):
I reproduced the existing bug in docker and verified the changes resolve it.

